PR #194, opened 2014-09-01
WIP: adding what's new to docs

drammock:
copied framework from `mne-python`... seems like there ought to be a way to automate at least some of this (like the contributors to each release, number of commits, etc).  Any ideas?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,20 +4,8 @@
 environment:
 
   matrix:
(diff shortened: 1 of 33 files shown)


drammock:
ugh.  I still don't have the hang of this rebasing business, apparently...  every time I do it it bloats my PRs with all those other commits that are already part of master.  I'm gonna close this one and re-open a new PR that has just the changes to the `whats_new` document.

PR #232, opened 2015-05-08
ENH: Add Cedrus box

Eric89GXL:
Ready for review/merge @rkmaddox @drammock

### DIFF
--- a/expyfun/_input_controllers.py
+++ b/expyfun/_input_controllers.py
@@ -457,12 +457,18 @@ class CedrusBox(Keyboard):
         self._dev = dev
         self._keyboard_buffer = []
         super(CedrusBox, self).__init__(ec, force_quit_keys)
+        ec._time_correction_maxs['keypress'] = 1e-3  # higher tolerance
 
     def _get_timebase(self):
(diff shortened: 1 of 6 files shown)


expyfun/_input_controllers.py
```diff
@@ -440,3 +440,56 @@ def _point_in_tri(self, pos, tri):
...
             return False
+
+
+class CedrusBox(Keyboard):
+    """Class for Cedrus response boxes
+
+    Note that experiments with Cedrus boxes are limited to ~4 hours due
+    to the data type of their counter (milliseconds since start as integers).
+    """
+    def __init__(self, ec, force_quit_keys):
```
rkmaddox: Four hours since experiment started or since Cedrus was plugged in? If the former, we need to compensate somehow, but I suspect the latter.

Eric89GXL: experiment started

rkmaddox: Cool. We may want to add a warning to the log file or something so that this knowledge gets passed on, but looks good for now.

PR #240, opened 2015-05-17
WIP: Barplot enhancements

drammock:
I've been using this code for many months (I used this branch when doing the recent JASA plots) so I think it's pretty safe, but I want to double-check that the tests are adequate before giving the go-ahead for merge.  Feel free to review the code changes now though.


Eric89GXL:
This all looks reasonable to me, and we can iterate in `master` if we find problems in use later

Eric89GXL:
Ahh crap, just noticed that it was WIP. Sorry about that... feel free to beef up tests in a subsequent PR if the current ones are inadequate.

PR #243, opened 2015-05-31
Better barplot tests

drammock:
@Eric89GXL, time to do your penance for merging a WIP pull request last week :)


expyfun/analyze/_viz.py
```diff
@@ -98,46 +98,49 @@ def barplot(h, axis=-1, ylim=None, err_bars=None, lines=False,
...
+        treated as raw height values. If `h` is a pandas ``DataFrame`` and
+        `bar_names` is ``None``, `bar_names` will be inferred from the
+        ``DataFrame``'s `column` labels (if ``axis=0``) or `index` labels.
     axis : int
         The axis along which to calculate mean values to determine bar heights.
-        Ignored if ``h`` is 0- or 1-dimensional.
+        Ignored if `h` is 0- or 1-dimensional.
     ylim : tuple | None
-        y-axis limits passed to ``matplotlib.pyplot.subplot.set_ylim()``.
+        y-axis limits passed to ``matplotlib.pyplot.subplot.set_ylim``.
```
Eric89GXL: do you know if we can do single-backticks for functions from other modules? I think it might actually work since sphinx should know about those other modules. But maybe it only works for numpy and scipy.

Eric89GXL:
Other than my minor gripe, LGTM

expyfun/analyze/_viz.py
drammock: don't know if it will actually link up, but the [numpy docstring standard](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#docstring-standard) does actually say

> Variable, module, function, and class names should be written between single back-ticks.

so I guess that means the references to pandas `DataFrame`s should get single backticks too?

drammock:
ok @Eric89GXL assuming a docstring change didn't bork the CIs, should be good to go.

Eric89GXL:
Thanks @drammock

PR #247, opened 2015-08-25
WIP, ENH: Add CRM corpus stimuli

rkmaddox:
Very much a work in progress--not sure the best way to include this in the main repo. Particularly what to do with the data. It should ship with sampling rates of 24414 and 44100, but I am not sure what to do when someone uses the prepare corpus functionality at a different sampling rate.

### DIFF
--- a/expyfun/stimuli/__init__.py
+++ b/expyfun/stimuli/__init__.py
@@ -9,6 +9,7 @@ from ._mls import (compute_mls_impulse_response, repeated_mls,
 from ._stimuli import rms, play_sound, window_edges
 from ._vocoder import vocode, get_band_freqs, get_bands, get_env, get_carriers
 from .._tdt_controller import get_tdt_rates
+from ._crm import crm_sentence, crm_response_menu
 
 # for backward compat (not great to do this...)
 from ..io import read_wav, write_wav
--- a/expyfun/stimuli/__init__.py
+++ b/expyfun/stimuli/__init__.py
@@ -9,7 +9,7 @@ from ._mls import (compute_mls_impulse_response, repeated_mls,
 from ._stimuli import rms, play_sound, window_edges
 from ._vocoder import vocode, get_band_freqs, get_bands, get_env, get_carriers
 from .._tdt_controller import get_tdt_rates
-from ._crm import crm_sentence, crm_response_menu
+from ._crm import crm_sentence, crm_response_menu, crm_prepare_corpus
 
 # for backward compat (not great to do this...)
 from ..io import read_wav, write_wav
--- a/expyfun/stimuli/_crm.py
+++ b/expyfun/stimuli/_crm.py
@@ -9,14 +9,14 @@ import scipy.signal as sig
 import expyfun.visual as vis
 
 fs_binary = 40e3
-sexes = {
+_sexes = {
     'male': 0,
     'female': 1,
     'm': 0,
     'f': 1,
     0: 0,
     1: 1}
-talker_nums = {
+_talker_nums = {
     '0': 0,
     '1': 1,
     '2': 2,
@@ -25,7 +25,7 @@ talker_nums = {
     1: 1,
     2: 2,
     3: 3}
-callsigns = {
+_callsigns = {
     'charlie': 0,
     'ringo': 1,
     'laker': 2,
@@ -50,7 +50,7 @@ callsigns = {
     5: 5,
     6: 6,
     7: 7}
-colors = {
+_colors = {
     'blue': 0,
     'red': 1,
     'white': 2,
@@ -63,7 +63,7 @@ colors = {
     1: 1,
     2: 2,
     3: 3}
-numbers = {
+_numbers = {
     'one': 0,
     'two': 1,
     'three': 2,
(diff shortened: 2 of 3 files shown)


rkmaddox:
@Eric89GXL would be good if we could gchat about this at some point. There are different ways a lot of this could be implemented. There is also a lot of cleaning up and error checking to add to make it generally usable.

rkmaddox:
Regarding data prep, maybe the best thing to do is only store the original raw binary files and have people prepare the corpus in the local experiment directory each time.

Eric89GXL:
How big is the complete corpus (MB/GB)? How long does it take to generate (minutes/hours)? Gotta find that download vs processing time tradeoff sweet spot. We can make the prepared/complete one (plus the script that made it from raw files) available if we need to.

rkmaddox:
The original corpus is 300 MB at 40 kHz. The 44100 corpus is 325 and took 28 minutes to make using 11 cores on my machine (it's a looooot of resampling), though that was using scipy's resample function and not expyfun's (which is private--should we have a public one?). So the 24414 corpus would be about 180 MB. Even if it takes an hour to process, it's set it and forget it, so it doesn't seem like it would be that big a hindrance. We could also put 44100 and 24414 versions on Lester and people could manually download those if they wanted to.

So I think I propose: only distribute the raw binary corpus with expyfun, and have people prepare (or download) their own resampled versions to be stored locally with each experiment. I am not a strong proponent of any plan, however.

Eric89GXL:
The resampling should actually be much faster if you use `expyfun`'s, since it *should* make use of `mne` which in turn can use CUDA.

I'm fine just putting the raw data up there, don't have a strong opinion either way.

drammock:
I don't think we should assume that all users are CUDA-savvy (e.g., might
be on laptop), and hence providing either (1) downloads at the two
different fs rates, and/or (2) a very clear readme that explains what to
expect from scipy vs expyfun/mne is needed.

The resampling should actually be much faster if you use expyfun's, since
it *should* make use of mne which in turn can use CUDA.

I'm fine just putting the raw data up there, don't have a strong opinion
either way.

—
Reply to this email directly or view it on GitHub
<https://github.com/LABSN/expyfun/pull/247#issuecomment-135128771>.

rkmaddox:
expyfun's resampling does not use CUDA. It took 28 minutes, exactly the same as scipy's resampling. But actually, parallelizing it by stimulus file, rather than doing one file with `n_jobs` > 1, is substantially faster anyway. So it's possible CUDA may speed it up, but I think it's more likely this is as fast as it's going to be until Eric implements polyphase resampling in scipy.

drammock:
I thought expyfun used mne for resampling (and therefore CUDA) when present
and properly configured?

rkmaddox:
Nope, it uses its own version that is based on mne (copied from an old version of mne?).

Eric89GXL:
@rkmaddox are you un-busy enough to pick this back up? (hahahaha)

rkmaddox:
Hahahaha. I think I left it in a state where it works, as in you could get
the CRM sounds you need. But I wanted it to get to the point where it was a
nice CRM class that got instantiated and handled everything for you.

Maybe the former is a good enough state to include, then we can keep that
functionality and add a nice CRM class alongside eventually?

--
Ross Maddox, Ph.D.
Postdoctoral Fellow
University of Washington
phone: 206-685-4662

Eric89GXL:
Yeah we can do it bit-by-bit, stimuli first, then paradigm codeblock later.

Eric89GXL:
Closing for now, we can resurrect once someone finds time to refactor, etc.

rkmaddox:
Sure. I am sure I (or my student) will be the one to reopen it. I also have the adaptive track code to put in here. Hopefully someone needs that before I do so they have to do it.

rkmaddox:
Well, that time has come and I finally need the CRM functionality. I assume new PR is the way to do it?

Will I need to rebase? I've never done that.

rkmaddox:
@Eric89GXL @drammock just in case you don't get pinged for comments on closed PRs.

Eric89GXL:
Yes ideally you would rebase. I'd recommend setting up "meld" as your diff
viewer before doing it.

If it goes horribly wrong, you can cut your losses and make a new branch
based on upstream master and copy paste your changes.

rkmaddox:
Hmm. I successfully rebased. Then I tried to push. It didn't work, and the forums told me to force push. So I did, but now I cannot reopen the PR because it was force pushed. Should I open a new PR?

rkmaddox:
Ah, from reading it seems like it might just be a privileges thing.

rkmaddox:
It also appears the rebase may not have gone right, since I now can't import ``expyfun.analyze``. I'll keep trying to figure it out.

rkmaddox:
FYI, I've rebased and gotten it to a point where I think it is sufficiently functional and documented. That has happened on my `rkmaddox:crm` branch, which is not showing up on this PR since it is closed.

drammock:
I can't seem to reopen it either... The button is greyed out.

rkmaddox:
Ok. I'm out today, but it seems like the thing to do is just open a new PR.

Issue #248, opened 2015-08-27
Should use mne/CUDA for resampling


drammock:

rkmaddox:
I'm not sure it should. It's a whole new dependency for one piece of functionality. We should at least confirm it's faster before we add the dependency.

Eric89GXL:
All hail the mighty CUDA:
```
import numpy as np
from mne.filter import resample
from time import time
import matplotlib.pyplot as plt

durs = (0.1, 0.5, 1., 2., 5., 10.)  # sec
n_durs = len(durs)
fs_in = 44100
fs_out = 24414
n_rep = 10
n_sig = 12
t = np.empty((3, n_durs))
kinds = (1, 6, 'cuda')
for ii, dur in enumerate(durs):
    n_samp = int(dur * fs_in)
    sig = np.empty((n_sig, n_samp))
    for ti, n_jobs in enumerate(kinds):
        resample(sig, 1, 10, n_jobs=n_jobs)
        t0 = time()
        for _ in range(n_rep):
            resample(sig, 24414, 44100, n_jobs=n_jobs)
        t[ti, ii] = (time() - t0) / n_rep

colors = ['k', (0.5, 0.5, 0.5), (0., 1., 0.)]
fig, ax = plt.subplots(1, 1)
for ki, (kind, color) in enumerate(zip(kinds, colors)):
    ax.plot(durs, t[ki], color=color, label=kind)
ax.set(xscale='log', yscale='log', title='mne-python resample()')
ax.set(xlabel='Signal duration (sec)', ylabel='Processing time (sec)')
ax.set(xticks=durs, xticklabels=durs)
plt.legend(title='n_jobs=', loc='SouthEast')
plt.show()
```
Note this is a log-log plot:
![figure_1](https://cloud.githubusercontent.com/assets/2365790/9531674/3b163c50-4cd5-11e5-828c-c4059793878f.png)

rkmaddox:
Oh my. So: require mne-python, or copy it's resample code into our own library?

Eric89GXL:
I think requiring mne-python is okay, it's a pure python dependency anyway, even installable using `pip`.

rkmaddox:
K. I'll do it.

drammock:
thx for volunteering @rkmaddox

drammock:
closed by #249

rkmaddox:
[...]

PR #249, opened 2015-08-28
MRG: use mne resample function

rkmaddox:
Addresses #248. Only difference is the mne `resample` has a `verbose` argument. Should I set that to `False` so it behaves as before, or not worry about it?


rkmaddox:
@drammock, in case you're not subscribe.

rkmaddox:
`make flake` causing tests to fail.

rkmaddox:
Correction: 
Lacking `future` module causing appveyor to fail.
`make flake` causing Travis to fail.

Eric89GXL:
WTF Pyglet

Can you change the CIs to `pip install pyglet` instead of installing the `tip.zip`? We shouldn't need the tip anymore.

expyfun/stimuli/_filter.py
```diff
@@ -7,118 +7,11 @@
...
-
-    # Restore the original array shape (modified for resampling)
-    y.shape = orig_shape[:-1] + (y.shape[1],)
-    if axis != orig_last_axis:
-        y = y.swapaxes(axis, orig_last_axis)
-
-    return y
-
-
 def _fft_resample(x, W, new_len, npad, to_remove):
```
Eric89GXL: we shouldn't need this, either

rkmaddox:
Changed the `pip` line and added `mne` as a dependency, but seems like `make flake` still causes a problem in Travis. Any thoughts?

rkmaddox:
Appveyor now happy. Travis still sad for aforementioned reason.

Eric89GXL:
You can combine those two pip install lines into one command, "pip install mne pyglet"

Do we still need _smart_pad? git grep it. Can you also check the other functions in that file? Not sure how many were related only to resampling. 

Get rid of unused imports to make Travis happy. You might want to enable pyflakes in spyder so you can see those while editing.

Eric89GXL:
You can also run make flake locally so you don't have to wait for Travis for feedback

rkmaddox:
Ah, I understand. I thought it was a problem with pyflake, not that it was catching problems with our code. Should be fixed now.

Eric89GXL:
Check coverage CI, `_smart_pad` isn't needed anymore.

Can you update the README.rst to say mne is required now? Although maybe we should actually only need it for stimulus generation, not e.g. presentation. So in that case, what we need to do is something like:

```
try:
    from mne.filter import resample as _resample
except ImportError:
    _resample = None


def resample(...)
    """..."""
    if _resample is None:
        raise ImportError(...)
    _resample(...)
```

Eric89GXL:
or you can nest the `mne` import within the `resample` function itself if you want

rkmaddox:
There is no `resample` function. It just imports the MNE one. However, I was thinking about changing that. `mne`'s `resample` assumes that outside of the signal are mirror-imaged versions of that signal, but it is more typical to assume zeros, and that is what the previous function did.

So I was considering making `expyfun`'s `resample` wrap `mne`'s. Seem like an OK plan?

rkmaddox:
Ohhhhhhhhh I just understood how the `resample` function was being created. Yikes. I didn't really process that code at the bottom.

Eric89GXL:
So you know what to do, or is it still unclear?

rkmaddox:
Is this sufficient?

```python
"""Stimulus resampling functions
"""

from distutils.version import LooseVersion

import mne
if LooseVersion(mne.__version__) < LooseVersion('0.8'):
    raise ImportError('mne-python too old')
from mne.filter import resample  # NOQA
```

rkmaddox:
That doesn't take care of the edge handling, though. Just not sure if it _should_ be taken care of. (I think it probably should, as this resample function will be used primarily for audio, rather than brain signals.)

.travis.yml
```diff
@@ -34,12 +34,8 @@ before_install:
...
     # Only install these deps if necessary (slow on Py3k)
     - travis_retry pip install -q nose-timer flake8;
-    - if [ "${PYTHON}" == "3.4" ]; then
-        travis_retry pip install -q pyglet;
-      else
-        travis_retry pip install -q http://bitbucket.org/pyglet/pyglet/get/tip.zip;
-      fi;
-    # Full dependencies (pandas, mne-python, h5py, joblib)
+    - travis_retry pip install -q pyglet mne;  # mne is a dependency now that resample is taken from it
+    # Full dependencies (pandas, mne-python, h5py, joblib, mne-python)
```
drammock: since you're mucking around in this file, note the duplicated "mne-python" in this comment

rkmaddox:
Never mind. It appears it pads with zeros, and I was remembering incorrectly. This should be good to go, I think, if I make those changes from the previous comment.

Eric89GXL:
I don't think `mne` should be a mandatory requirement for `import expyfun`, but it can be a mandatory requirement for `from expyfun.stimuli import resample; resample(...)`, make sense? See my comment above for suggested mechanism.

rkmaddox:
Ah, yes. I understand now.

rkmaddox:
Is the call to LooseVersion still necessary, or will it also cause an import error if someone tries to import `resample` from `mne.filter` in an older install?

Eric89GXL:
I wouldn't worry about it in this case, 0.8 is over a year old now and I doubt any of our user base is using an old version. But yeah, I suspect that a pre-0.8 will either give an `ImportError` or maybe some other error when you go to actually use it, which is fine.

rkmaddox:
I think this should be good to go, but take a look at the docstring thinger. I wanted it to stay current with the function, but not sure if what I did was kosher.

rkmaddox:
Never mind. The way I did it it doesn't show the arguments. I'll just do it the less elegant but probably more functional way.

Eric89GXL:
Yeah, unfortunately the `args/kwargs` + docstring is something that I haven't found a way around violating DRY to get right in these situations

rkmaddox:
The `__doc__` trick worked OK. But I did not see a way to get the args to work.

drammock:
LMK when this is ready for an extra set of eyes and a `make test` on my machine

rkmaddox:
Hmm. How about this? Seems to work on my machine.

```python
def _resample_error(*args, **kwargs):
    """mne-python is required to use the resample function
    """
    raise ImportError('mne-python is required to use the resample function')

try:
    from mne.filter import resample
except ImportError:
    resample = _resample_error
```

rkmaddox:
@drammock ready for test. Changing to MRG.

Eric89GXL:
LGTM, +1 for merge. @rkmaddox can you look quickly at why our coverage went down? Maybe we just removed a well-covered file, which is what I suspect happened.

rkmaddox:
Yep, it's because the number of covered lines in `_filter.py` went down by ~60.

rkmaddox:
That's what we get for DRY.

drammock:
weird, the final commit 57363048775ce862065c3d67763f6c9756d9abb6 passed all tests, but the homepage is showing [appveyor failure](https://ci.appveyor.com/project/Eric89GXL/expyfun/branch/master), failing on "Testing EC with TDT" line.

Eric89GXL:
I set it to rebuild, might have just been a fluke

Eric89GXL:
all better

drammock:
thanks @Eric89GXL

Issue #182, opened 2014-08-18
FIX: Text display on windows with larger font sizes


rkmaddox:
So the recent addition of font control in `screen_prompt` has revealed what I anticipate to be a nasty Pyglet bug. Displayed text has random (as far as I can tell) but reproducible spacing issues where large amounts of horizontal space are placed in between letters in the string. This is happening post-wrap, as it causes the text to go off the screen to the right instead of creating more lines.

Problem gets worse with increasing text size (wanted to use 40). Workaround was to use a smaller text size (seems to be OK with 30). This will need to be fixed, as some subjects have expressed difficulty reading instructions at the default size of 24 pts.

PR #252, opened 2015-09-13
ENH: Use GLSL pipeline for efficient draws

### LINKED ISSUE #251: FIX: about 4% of flips are happening late
@lkishline and I are getting an AV experiment set up, which involves some carefully timed flashes. We are timing these flips such that there is a first flip when `start_stimulus` is called and all other flips use the `when` argument relative to that one. We have set it up such that the time we pass for `when` is equal to the initial flip time + a time corresponding to a whole number of frames - `0.9 / frame_rate`. This should very reliably put the flips where we want them.

Instead, we are losing about 10% of trials because some trials have late flips, sometimes by even more than one frame. @Eric89GXL @drammock, do you guys have any idea how the flip timing could be made more reliable?

Eric89GXL:
Closes #251.

@drammock @rkmaddox @lkishline can you try this branch to see if you get better results?

### DIFF
--- a/expyfun/_input_controllers.py
+++ b/expyfun/_input_controllers.py
@@ -421,9 +421,9 @@ class Mouse(object):
     def _point_in_tris(self, pos, obj):
         """Check to see if a point is in any of the triangles
         """
-        these_tris = obj._tris.reshape(-1, 3)
+        these_tris = obj._tris['fill'].reshape(-1, 3)
         for tri in these_tris:
-            if self._point_in_tri(pos, obj._points[tri]):
+            if self._point_in_tri(pos, obj._points['fill'][tri]):
                 return True
         return False
 
--- a/expyfun/_input_controllers.py
+++ b/expyfun/_input_controllers.py
@@ -421,15 +421,11 @@ class Mouse(object):
     def _point_in_tris(self, pos, obj):
         """Check to see if a point is in any of the triangles
         """
-        found = False
-        index = 0
-        while index < len(obj._tris) and not found:
-            points = obj._tris[index:index + 3]
-            tri = np.array([[obj._points[2 * p], obj._points[2 * p + 1]] for
-                            p in points])
-            found = self._point_in_tri(pos, tri)
-            index += 3
-        return found
+        these_tris = obj._tris.reshape(-1, 3)
+        for tri in these_tris:
+            if self._point_in_tri(pos, obj._points[tri]):
+                return True
+        return False
 
     def _point_in_tri(self, pos, tri):
         """Check to see if a point is in a single triangle
--- a/expyfun/visual/_visual.py
+++ b/expyfun/visual/_visual.py
@@ -7,6 +7,10 @@
 
 # make RawImage work
 
+from ctypes import (cast, pointer, POINTER, create_string_buffer, c_char,
+                    c_int, c_float)
+from functools import partial
+
 import numpy as np
 from matplotlib.colors import colorConverter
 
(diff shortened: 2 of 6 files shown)


Eric89GXL:
Here is current `master` for me:
master

And on this branch:
glsl

Note the y-axes are different. So now 100 and 500 are both fast and reliable, and 500 takes ~400 microseconds to draw on my machine. Hopefully you see something similar.

Eric89GXL:
I'll need to make some tweaks to Appveyor to get it to build and work properly there, but at least Travis is happy and tests pass locally for now.

drammock:
Thanks @Eric89GXL. Will test tomorrow. since circles default to 200 edges
this should hopefully solve the current problem. I'm still puzzled that you
see a few long delays in the plot for small numbers of edges though... Any
speculations as to why? If you generate the plot again is it deterministic?

rkmaddox:
Works very well on my computer. @lkishline will post test results from the lab computer imminently.

Thanks for the fix.

expyfun/_input_controllers.py
```diff
@@ -421,15 +421,11 @@ def _point_in_object(self, pos, obj):
...
-            tri = np.array([[obj._points[2 * p], obj._points[2 * p + 1]] for
-                            p in points])
-            found = self._point_in_tri(pos, tri)
-            index += 3
-        return found
+        these_tris = obj._tris['fill'].reshape(-1, 3)
+        for tri in these_tris:
+            if self._point_in_tri(pos, obj._points['fill'][tri]):
+                return True
+        return False
```
rkmaddox: Nice change. Much cleaner.

lkishline:
Results on MARLO, not so hot:

drawtest

lkishline:
Here is the same graph, except with `audio_controller='pyglet', response_device='keyboard', trigger_controller='dummy'` passed to `ExperimentController`:

drawtest-pyglet

!!!!!

rkmaddox:
So Lindsey and I ran it some more, and it basically works pretty darn well, regardless of the settings. I have no idea why you guys previously found a difference based on whether TDT was in use or not, but there no longer seems to be one. I have no ideas on what might be controlling it. @lkishline, did you ever run it on a bunch more blocks to see if it was still perfect?

lkishline:
@rkmaddox was just about to run it on 4 blocks and got a frame rate error on our assertion line....not sure why...But tried to run it again and it started up just fine. It is currently in the middle of running 4 blocks.

lkishline:
Alright, well I ran it on 4 blocks and only saw one missed frame!

Although one of the fixes made the fixation dot look different and text between my blocks disappear. But it could be cause I'm using a different version of expyfun then what I tested my script on.

rkmaddox:
Good that there are far fewer missed frames. Losing 1 in 400 trials seems like quite acceptable attrition.

Don't know why the fixation dot would look different. Does the fixation dot with default parameters look right? On my linux box it looks normal using this branch.

lkishline:
Oh wait - never mind. I was using the ConcentricCircles for the fixation dot and the answer cue dot because I was moving them around and changing the colors/ sizes.

drammock:
@lkishline is the text issue similar to what is described in #182, or is the text just simply gone?

lkishline:
@drammock I don't think so? screen_prompt text within in my block/trials loop looks like this. Just two dots in the middle of the screen. And when I hit the button it turns to one dot, and then I hit the button to continue again and it goes into the trial. The text before the block loop is fine.
image

drammock:
that is definitely something different.

lkishline:
Yes, I thought I would take it upon myself to create as many mysterious issues as possible. So far I think I'm doing pretty well on that endeavor.

Eric89GXL:
Can you share a minimal example recreating the behavior?

lkishline:
@Eric89GXL Yes, I will give that a go. Give me a bit.

lkishline:
Okay, so hopefully all this gist stuff works....[gist] (https://gist.github.com/lkishline/3d7f35b0ca428caadc3f)

Eric89GXL:
Do there need to be multiple blocks to see the problem? Multiple trials? In other words, is that example really minimal to reproduce the issue?

lkishline:
@Eric89GXL ah, right. let me go back and test.

Eric89GXL:
Cool -- the "minimal" idea is important because it makes a pretty good step toward isolating the problem to a small set of calls, which is why I'm being picky about it.

lkishline:
Yea, I just realized that by not making it truly minimal I was asking others to do more work. My bad!

lkishline:
[Updated!] (https://gist.github.com/lkishline/3d7f35b0ca428caadc3f)

drammock:
@lkishline's second gist works normally for me on Linux (moctezuma).  Will try to look at it on marlo tomorrow morning.

Eric89GXL:
I can replicate on Windows, fix coming

Eric89GXL:
@lkishline updated, example works over here now. Feel free to give it a try. FYI this is the example after a couple other minor tweaks to facilitate testing:
```
from expyfun import ExperimentController
from expyfun.visual import ConcentricCircles

with ExperimentController('Test', full_screen=False, participant='foo',
                          session='bar', output_dir=None) as ec:
    ec.screen_prompt('a', max_wait=0.3)
    dot = ConcentricCircles(ec)
    dot.draw()
    ec.flip()
    ec.wait_secs(0.3)
    ec.screen_prompt('b', max_wait=0.3)
```

lkishline:
@Eric89GXL It works! Thanks so much. 

And I'll work on creating better minimal examples in the future.

drammock:
I'm guessing @rkmaddox is the one of us best-qualified to vet your OpenGL code, @Eric89GXL, so I'll leave the merge decision to him.  Otherwise LGTM.

rkmaddox:
Yikes. I have zero GL experience. Will do my best, though.

Eric89GXL:
In case you want to learn a bit of background:

http://duriansoftware.com/joe/An-intro-to-modern-OpenGL.-Chapter-1:-The-Graphics-Pipeline.html

expyfun/visual/_visual.py
```diff
@@ -176,27 +306,39 @@ def set_line_width(self, line_width):
...
+                                self._buffers[kind]['array'])
+                loc_pos = gl.glGetAttribLocation(self._program, b'a_position')
+                gl.glEnableVertexAttribArray(loc_pos)
+                gl.glVertexAttribPointer(loc_pos, 2, gl.GL_FLOAT, gl.GL_FALSE,
+                                         0, 0)
+                loc_col = gl.glGetUniformLocation(self._program, b'u_color')
+                gl.glUniform4f(loc_col, *self._colors[kind])
+                cmd()
+                # The following line is probably only necessary because
+                # Pyglet makes some ossumptions about the GL state that
```
rkmaddox: typo (assumptions)

rkmaddox:
Other than a single typo in the comments, I think it looks good (but admittedly, much of it was not understandable to me). I think we should watch our stimuli carefully as we develop the next couple experiments to make sure things are really working. And hey, maybe this will fix the text spacing bug?

+1 for merge.

Eric89GXL:
This doesn't touch TextVisual at all unfortunately, so it shouldn't affect that bug. I still saw it while debugging :(

Typo fixed.

rkmaddox:
Thanks for the vast improvement.

PR #255, opened 2015-10-15
ENH: allows download_version from expyfun forks

drammock:
tested locally, passes nose.  Ready for merge as far as I'm concerned...  @Eric89GXL  @rkmaddox are you cool with this?

### DIFF
--- a/expyfun/_git.py
+++ b/expyfun/_git.py
@@ -86,7 +86,7 @@ def download_version(version='current', dest_dir=None, fork='LABSN'):
     orig_stdout = sys.stdout
     try:
         from setup import git_version, setup_package
-        assert git_version().lower() == version[:7].lower()
+        assert git_version()[0].lower() == version[:7].lower()
         sys.stdout = StringIO()
         with warnings.catch_warnings(record=True):  # PEP440
             setup_package(script_args=['build', '--build-purelib', dest_dir])
--- a/expyfun/_git.py
+++ b/expyfun/_git.py
@@ -9,6 +9,11 @@ from ._version import __version__
 
 this_version = __version__[-7:]
 
+try:
+    from ._version import __fork__
+    this_fork = __fork__
+except ImportError:
+    this_fork = None
 
 try:
     run_subprocess(['git', '--help'])
@@ -108,6 +113,11 @@ def assert_version(version):
         Version to check (7 characters).
     """
     _check_version_format(version)
+    if this_fork is not None and this_fork.lower() != 'labsn':
+        raise RuntimeWarning('You are running a forked version of the expyfun '
+                             'codebase (fork "{}"). Do not do this if you are '
+                             'running an actual experiment; only do this '
+                             'during development.'.format(this_fork))
     if this_version.lower() != version.lower():
         raise AssertionError('Requested version {0} does not match current '
                              'version {1}'.format(version, this_version))
--- a/expyfun/_git.py
+++ b/expyfun/_git.py
@@ -36,7 +36,7 @@ def _active_version(wd):
     return run_subprocess(['git', 'rev-parse', 'HEAD'], cwd=wd)[0][:7]
 
 
-def download_version(version='current', dest_dir=None):
+def download_version(version='current', dest_dir=None, fork='LABSN'):
     """Download specific expyfun version
 
     Parameters
@@ -48,6 +48,8 @@ def download_version(version='current', dest_dir=None):
     dest_dir : str | None
         Destination directory. If None, the current working
         directory is used.
+    fork : str
+        GitHub username of the expyfun fork to download from.
 
     Notes
     -----
@@ -67,7 +69,7 @@ def download_version(version='current', dest_dir=None):
     # fetch locally and get the proper version
     tempdir = _TempDir()
     expyfun_dir = op.join(tempdir, 'expyfun')  # git will auto-create this dir
-    repo_url = 'git://github.com/LABSN/expyfun.git'
+    repo_url = 'git://github.com/{}/expyfun.git'.format(fork)
     run_subprocess(['git', 'clone', repo_url, expyfun_dir])
     version = _active_version(expyfun_dir) if version == 'current' else version
     try:
@@ -92,8 +94,9 @@ def download_version(version='current', dest_dir=None):
         sys.stdout = orig_stdout
         sys.path.pop(sys.path.index(expyfun_dir))
         os.chdir(orig_dir)
-    print('Successfully checked out expyfun version:\n\n%s\n\ninto '
-          'destination directory:\n\n%s\n' % (version, op.join(dest_dir)))
+    print('\n'.join(['Successfully checked out expyfun version:', version,
+                     'from the "{}" fork of expyfun'.format(fork),
+                     'into destination directory:', op.join(dest_dir)]))
 
 
 def assert_version(version):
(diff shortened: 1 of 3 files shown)


Eric89GXL:
I'd rather not. It seems to defeat the purpose of having things linked to the `master` branch. My assumption is that people will comment out `assert_version` until they're ready to go. Do you need it for something?

Eric89GXL:
(i.e., code should *really* be in `LABSN/expyfun.git` to be deployed for an experiment)

Eric89GXL:
Ohhh... *download_version*, I was thinking of `assert_version` (although it probably doesn't really matter). Yeah I'm okay with it then :)

Eric89GXL:
I'll let @rkmaddox merge if he's happy

drammock:
I totally agree that we should never run exps off of forks. you're right
that this is strictly for dev purposes, but I've missed this feature 2 or 3
times now during dev testing so I think others may also find it useful.

rkmaddox:
So if someone runs `assert_version` using a hash from the non-LABSN fork, does that mean they can run the experiment off of that fork--plan is to trust them not to do that? (That seems fine, just want to confirm.)

Eric89GXL:
I think so, yeah. But people should be strongly encouraged to use the
version= kwarg from the EC constructor instead of the function where it's
not available, so I expect only experts (Dan) to use the fork feature.

Eric89GXL:
Never mind, got the download and assert steps confused again. Perhaps it's
bed time here.

rkmaddox:
I'll let Dan chime in. Good night.

drammock:
It is correct to say that, if someone did `download_version(some_commit_hash, fork='drammock')` and then subsequently did `assert_version(some_commit_hash)` in a script in the same directory, the assertion would hold.  Also would hold for the version argument of `ExperimentController`, which uses `assert_version` under the hood.  Is this too lax?  I'm willing to make it harder, like making a private function called `_download_version_from_fork` or similar...

rkmaddox:
It's probably fine. I'd also be ok with a warning when `assert_version` is satisfied with a hash that is from a fork. Is there sufficient information in the `expyfun` directory (git metadata, presumably) do tell if it is or isn't?

drammock:
Not currently; `download_version` doesn't actually clone the repo, so the `.git` folder is not present.  The commit hash is stored in `_version.py`.  I can try to get the repo name in there too, so that we could query it.  Stand by.

drammock:
looks like I can do this through the `write_version` function in `setup.py`, but it will have to wait until tomorrow.

drammock:
OK @Eric89GXL I'm going to need to ask for help here.  `make test` passes locally, but Travis is failing in ways that I don't understand.  The jobs tagged as python 2.7 fail on `test_resampling` due to not finding mne python. The job tagged `python 3.4 DEPS=full TEST=standard _EXPYFUN_SILENT=true` seems to be actually running on python 2.7, not 3.4, and subsequent travis jobs give two bizarrely inconsistent errors, [this one](https://travis-ci.org/LABSN/expyfun/jobs/87773136#L802) corresponding to [this commit](https://github.com/drammock/expyfun/commit/4eded9ba9013c07cf488419a8ab8952b0f7003ec) and [this one](https://travis-ci.org/LABSN/expyfun/jobs/87778990#L308) corresponding to [this commit](https://github.com/drammock/expyfun/commit/7d796ce552b93108fe2c13726d849e2f0c368bc0).

Can you offer any insight?

Eric89GXL:
@drammock I think the 3.4 jobs are running 3.4. If you think they're running 2.7 because of the column on Travis that's a red herring, because we don't actually use the system Python for 3.4 but instead Anaconda. You can also tell from the log because you get silly warnings like this on the 3.4 build:
```
/home/travis/anaconda/envs/testenv/lib/python3.4/site-packages/matplotlib/collections.py:590: FutureWarning ...
```
The resampling problem should be fixed by #261, which I'm going to merge so I can restart your tests and hopefully you'll not see them in this PR anymore. I'll look into the other error(s) shortly.

Eric89GXL:
BTW the way I typically debug these things is to actually use a `miniconda` install a lot like what Travis does. It makes it *much* easier to reproduce and debug errors that I see on Travis.

Eric89GXL:
Argh, looks like you might need a rebase to benefit from #261. I thought Travis tested merges, but maybe because your last commit was before that PR was merged it still merges with the old master...?

setup.py
```diff
@@ -44,16 +44,32 @@ def _minimal_ext_cmd(cmd):
...
         except OSError:
             GIT_REVISION = "Unknown"
+        try:
+            cmd = ['git', 'config', '--get', 'remote.origin.url']
+            out = _minimal_ext_cmd(cmd).decode('utf-8').strip()
+            start_idx = 0
+            for prefix in [b'git@github.com:', b'git://github.com/',
+                           b'http://github.com/', b'https://github.com/',
+                           b'ssh://git@github.com/']:
+                if out.startswith(prefix):
```
Eric89GXL: At this point, `out` has been decoded and is thus a string, and you're using a bunch of `bytes` instances to compare against. I think changing these to not have the `b` in front should fix it. You need to make sure they're either both `bytes` or both `string`, and you should be good..

drammock: That's what I thought too, but see Travis build 911.3 line 802 (the first of the two I linked to in my plea for help). It ran with the decode statement and without the `b`s and still errored ("first arg must be byte or tuple of bytes, not str", i.e. the opposite error from build 912.3).

Eric89GXL: I'd put in the code we both expect to work, but add a `print((type(out), type(prefix), GIT_REVISION)` in a commit, and see what you get (after rebase to get rid of 2.7 errors). I wonder if there is some weirdness with the version checkouts. But that information should help us debug it.

drammock: Will do (in an hour or so, when I'm not typing on a phone). Thanks for the quick fix on the mne import issue BTW

expyfun/_git.py
```diff
@@ -105,6 +113,11 @@ def assert_version(version):
         Version to check (7 characters).
     """
     _check_version_format(version)
+    if this_fork is not None and this_fork.lower() != 'labsn':
+        warnings.warn('You are running a forked version of the expyfun '
+                      'codebase (fork "{}"). Do not do this if you are '
+                      'running an actual experiment; only do this during '
+                      'development.'.format(this_fork), RuntimeWarning)
```
drammock: I'm not sure how to test this, because the tests run within the system's expyfun directory.  I can download expyfun to a `_TempDir()` and I can `sys.path.insert` that dir at position zero, but the cwd will always get checked first before the rest of the path options, so I can't import anything from the forked-and-downloaded version of the codebase, hence I cannot test (within the unittests) whether this warning actually comes up.  Am I missing something, or do we have to live with this decline in coverage?

drammock:
ok @Eric89GXL, Travis is happy, and (assuming AppVeyor eventually comes to the party) other than my most recent line comment I think this is ready for review.

Eric89GXL:
Did it end up being a commit-order thing? Like you needed to make a commit, then make *another* commit to refer to that one to get the useful changes from it? Now I remember there being something like that.

drammock:
I think that's what happened, but honestly when Travis stopped complaining
I stopped digging for the cause.

Eric89GXL:
Thanks @drammock for figuring this out

PR #258, opened 2015-10-16
ENH, FIX: adds mandatory version string argument to EC

### LINKED ISSUE #253: ENH: more stringent version checking
Propose adding a new, mandatory argument to `ExperimentController` called `expyfun_version` (or maybe just `version`?), that accepts a string (7-character GitHub commit hash) or `None`.  This argument would have no default, so users would be **required** to provide it, thus making the change non-backwards-compatible.  On init, the existing `expyfun.assert_version()` would be used to check the version string; specifying `None` would simply not run the check.  The ability to use `assert_version` *outside* of an `ExperimentController` instance would remain, so that it could be used in scripts that only do stimulus generation, plotting, etc.

@rkmaddox and I just discussed this, want to run it by you @Eric89GXL before we start hacking.

### LINKED ISSUE #257: screen_prompt not working in current master
on MARLO

```python
from expyfun import ExperimentController
with ExperimentController(**kwargs) as ec:
    ec.wait_secs(1)
    # ec.flip()
    ec.screen_prompt('foo', wrap=False, max_wait=3)
    ec.screen_prompt('bar', wrap=False, max_wait=3)
```

result is that "foo" appears on screen for about the duration of 1 frame, then black screen for ~3 seconds, then "bar" for 3 seconds.  If the `ec.flip()` line is *not* commented out, `screen_prompt` behaves normally.  I think this issue came to light once before, but I don't remember the resolution (if there was one)...

drammock:
also adds initial flip to end of EC init.  Closes #257, closes #253

### DIFF
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -27,6 +27,7 @@ from ._sound_controllers import PygletSoundController, SoundPlayer
 from ._input_controllers import Keyboard, CedrusBox, Mouse
 from .stimuli._filter import resample
 from .visual import Text, Rectangle, _convert_color
+from ._git import assert_version
 
 
 # Note: ec._trial_progress has three values:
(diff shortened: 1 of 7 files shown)


Eric89GXL:
LGTM, ready from your end?

drammock:
Yep.

Eric89GXL:
Thanks @drammock

PR #260, opened 2015-10-16
FIX: Fix background color

### LINKED ISSUE #259: BUG: ec.set_background_color no longer accepts all color formats
Recent GL-related changes to expyfun have partially broken the ability to set background colors. I've created [a gist](https://gist.github.com/drammock/1645b12fdf4edd9ccdb4) that illustrates the problem.

ec0b487 is last known good
937ac4e setting bgcolor works, but has other major issues
771f9f6 setting bgcolor works only for strings like `'k'` or `'red'`; values like `'0.5'` and `np.array([0.5, 0.5, 0.5])` yield a white background.

@Eric89GXL can you spot a likely culprit in the changes you made in 771f9f6 ?  This is a show-stopper for `find_pupil_dynamic_range` which I need to run a few more subjs to address reviewer comments...  for the time being I'll work from last known good, but we need to keep those GL changes in there for @lkishline's stuff to work.

Eric89GXL:
Closes #259.

@drammock see if it works okay for you, if so feel free to merge

### DIFF
--- a/examples/experiments/drawing_methods.py
+++ b/examples/experiments/drawing_methods.py
@@ -20,7 +20,7 @@ print(__doc__)
 
 with ExperimentController('test', session='1', participant='2',
                           full_screen=False, window_size=[600, 600],
-                          output_dir=None) as ec:
+                          output_dir=None, version='dev') as ec:
     ec.screen_text('hello')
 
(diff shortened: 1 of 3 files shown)


drammock:
works, thanks!  Just waiting for AppVeyor

PR #261, opened 2015-10-28
FIX: Fix Travis

Eric89GXL:
I suspect this might fix Travis.


Eric89GXL:
I merged this 4-character fix to our CIs so that I could more easily debug @drammock's issue. Sorry for breaking standard practice...

PR #262, opened 2015-12-07
MRG: add circle

rkmaddox:
Moves text to side and makes an empty white circle in the middle so that the photodiode timing can be referenced exactly to the center of the screen.

### DIFF
--- a/examples/experiments/sync_test.py
+++ b/examples/experiments/sync_test.py
@@ -2,9 +2,7 @@
 =============
 A-V sync test
 =============
-
 This example tests synchronization between the screen and the audio playback.
-
 NOTE: On Linux (w/NVIDIA), XFCE has been observed to give consistent timings,
 whereas Compiz WMs did not (doubled timings).
 """
@@ -16,6 +14,7 @@ import numpy as np
 import matplotlib.pyplot as plt
 
 from expyfun import ExperimentController
+from expyfun.visual import Circle
 import expyfun.analyze as ea
 
 print(__doc__)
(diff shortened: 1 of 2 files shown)


rkmaddox:
Looks like `test_barplot` is angry with python 3. Will look into it.

rkmaddox:
Hmm. It passes in python 2 for Travis, but fails in 3. It fails in python 2 for appveyor. Error is the same when it fails. It passes on my machine, so I am having trouble figuring out how to fix. @drammock or @Eric89GXL any thoughts?

drammock:
The bug seems to be related to the saving functionality in the matplotlib backend; we test `['eps', 'pdf', 'png', 'raw', 'svg']` and maybe one or more of those doesn't work on the CI servers, or is just unstable? I hate to reduce our test coverage of the different export types, but maybe we should reduce the output types we test, or even kill the `fname` argument altogether.

@Eric89GXL by the way, can you tell me your trick for "re-starting" the CI builds when you think it's a bogus failure?

Eric89GXL:
Probably new mpl bug in conda. Yay!

There is a restart icon on the page for the build on travis. I'll look
deeper later but for now it's fine by me to cut out a renderer or two if
necessary.

drammock:
I restarted the Travis job and it failed again on the same line.

drammock:
Both of the failed cases are loading MPL 1.5.0 via conda; the successful py2k+linux builds are both loading python-matplotlib_1.1.1~rc1+git20120423-0ubuntu1_amd64.deb via apt-get.  So either the matplotlib API has changed in a non-backward-compatible way, or there's been a bug introduced as @Eric89GXL suggests.

Eric89GXL:
We could force conda to install mpl 1.4 if we want (see mne Python code),
I'm fine with that or reducing types to test

drammock:
@rkmaddox and I discussed this the other day, and the decision was to just remove all but the PDF file type from the testing suite.  If PDFs turn out to be the problem, then we'll revisit and try to come up with another solution.

drammock:
PR'd into @rkmaddox's PR... the ball is in thy court.

rkmaddox:
Dan fixed it. @Eric89GXL ready for merge. @drammock, thanks for the fix.

Eric89GXL:
Thanks @rkmaddox and @drammock

PR #264, opened 2016-01-27
MRG: better config

rkmaddox:


Eric89GXL:
let me know when I should review with a `MRG: ` prefix to the title and a ping (preferably after the CIs are happy, assuming they're bombing due to your changes)

rkmaddox:
@Eric89GXL @drammock Ready for review and merge.

examples/experiments/level_test.py
```diff
@@ -3,7 +3,7 @@
 Sound level test and visual size calibration
 ============================================
 
-This example tests the audio level and video size. For audio, it produces a 65
+This example tests the audio level and video size. For audio, it produces an 80
```
Eric89GXL: Just curious why we want to bump it up?

expyfun/_utils.py
```diff
@@ -571,6 +572,8 @@ def get_config(key, default=None, raise_error=False):
     else:
         with open(config_path, 'r') as fid:
             config = json.load(fid)
+            if key is None:
+                return config
```
Eric89GXL: this doesn't need to be indented (keep as little code in the context manager as necessary)

expyfun/_version.py
```diff
@@ -1 +1,2 @@
 __version__ = '2.0.0.dev0'
+__fork__ = 'rkmaddox'
```
Eric89GXL: whoops

Eric89GXL:
Other than my minor comments LGTM

examples/experiments/level_test.py
rkmaddox: More signal relative to noise floor (sound level meter and room noise). Seemed like the dB reading was more stable when testing in 226 yesterday.

rkmaddox:
Comments addressed. Will ping once tests are happy.

drammock:
two nosetest failures when running this branch locally (in `test_fork_warning` and `test_version`).  Looking into it now.

expyfun/_utils.py
```diff
@@ -592,13 +595,16 @@ def set_config(key, value):
...
+    key : str | None
+        The preference key to set. If None, a tuple of the valid
+        keys is returned, and ``value`` is ignored.
     value : str |  None
         The value to assign to the preference key. If None, the key is
         deleted.
     """
 
+    if key is None:
+        return known_config_types
```
Eric89GXL: now that I think about it we might as well return `sorted(known_config_types)` to be a little bit nicer to users

Eric89GXL:
Heh, you named your latest commit 'final fixes`. The optimism must be strong in you

rkmaddox:
Let's call it what it is: naivete. I can sort the thing. @drammock is looking to version test weirdness. Hold off on merge.

Eric89GXL:
ALL CAPS MUST BE RESPECTED

drammock:
One nosetest was my fault, due to an untracked test file (`test_fork_warning`) lingering in my local tree.  The other is, I think, due to changes @rkmaddox made to `_version.py`.  [This change](https://github.com/LABSN/expyfun/pull/255/files#diff-2eeaed663bd0d25b7e608891384b7298R72) in #255 added a second line to `_version.py` that has been removed.

This is actually a sticky problem for expyfun developers, but may not affect end users.  Changes are made automatically to `_version.py` when `setup.py` is run, but `setup.py` is not run every time we change branches because we all ran `setup.py develop` long ago.  So I think we may need to run the code that updates `_version.py` as part of the unittests, which argues for pulling it out of `setup.py` and into a separate file that gets called from `setup.py` and also from `make test`.  thoughts?

Eric89GXL:
Could we just make it so any modifications to `_version.py` are done in a `try:` block with a `finally:` to revert any changes?

Eric89GXL:
(within the part of the code in `setup.py` that does the mods)

drammock:
> Could we just make it so any modifications to _version.py are done in a try: block with a finally: to revert any changes?

I don't think so, because if we do that we won't get the correct version hash in there, which defeats our ability to `assert_version` and to require an explicit version string when instantiating `ExperimentController`.  Or am I mistaken about that?

drammock:
I think I was wrong before.  Here is my attempt to summarize how things work:  

- The version string in `_version.py` starts out as `2.0.0.dev0` and the commit hash only gets added when `setup.py` is run.  
- Our `download_version` code uses `setup.setup_package` to "install" the downloaded version (using `setup.py`) to `dest_dir` (typically the experiment's working directory), so version checking seems to work there.
- In `test_version.py`, we [clone drammock/expyfun](https://github.com/LABSN/expyfun/blob/master/expyfun/tests/test_version.py#L30) which is done in `_git.py` [here](https://github.com/LABSN/expyfun/blob/master/expyfun/_git.py#L78), then [checkout the requested hash](https://github.com/LABSN/expyfun/blob/master/expyfun/_git.py#L81), then ["install" it into a specific temporary directory](https://github.com/LABSN/expyfun/blob/master/expyfun/_git.py#L97).  Tests built into `_git.py`'s `download_version()` code suggest that all of this is working: in the test output below, the "successfully checked out expyfun" message gives the expected, requested hash.  However, when we later [parse the installed `_version.py` file](https://github.com/LABSN/expyfun/blob/master/expyfun/tests/test_version.py#L34-L37), it yields the hash of a different commit (but strangely, not even the most recent one in drammock/master: it's two commits later in the log from the one we requested).  Here is the error text from `make test`:

```sh
======================================================================
FAIL: Test version assertions
----------------------------------------------------------------------
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/drmccloy/Builds/expyfun/expyfun/tests/test_version.py", line 37, in test_version
    assert_equal(line1.split(' = ')[1][-8:-1], v)
AssertionError: 'f593bd4' != 'b798e65'
-------------------- >> begin captured stdout << ---------------------
Successfully checked out expyfun version:
b798e65
from the "drammock" fork of expyfun
into destination directory:
/tmp/tmpD39bZ7

--------------------- >> end captured stdout << ----------------------
    "'f593bd4' != 'b798e65'" = '%s != %s' % (safe_repr('f593bd4'), safe_repr('b798e65'))
    "'f593bd4' != 'b798e65'" = self._formatMessage("'f593bd4' != 'b798e65'", "'f593bd4' != 'b798e65'")
>>  raise self.failureException("'f593bd4' != 'b798e65'")
```

any ideas what the heck is going on here?

drammock:
d'oh!  If I had bothered to re-run `make test` **after** I had removed the lingering untracked test file, I might have realized that the untracked file was causing **both** problems.  Sorry to spam this PR with a bunch of pseudo-problems.  OK to merge.

Eric89GXL:
Hah! Good to know it was local

Eric89GXL:
Thanks @rkmaddox

PR #266, opened 2016-02-16
FIX: Bigger buffer and auto-stop

### LINKED ISSUE #278: ENH: make ec._playing more accurate via polling
motivated by changes made in #266, the hard requirement to call `ec.stop()` in between subsequent calls to `ec.play()` (and the corresponding `ec._playing` property that kept track of this) has been called into question.  However, as @Eric89GXL notes:

> we probably don't want people doing `.load_buffer` before the stimulus is done playing, and we probably don't want them doing that before `.trial_ok()` is called -- and we need `ec._playing` for that.

Eric89GXL:
This PR:

1. Doubles our buffer size to 4e6
2. Reduces circuit complexity
3. Reduces CPU load ~80% -> ~70%
4. Automatically stops+resets the buffer once playback completes

Todo:

* [x] Add checking of `load_buffer` size <= allowed size.
* [x] Check it runs on RP2 at UW.
* [x] Check triggers on RP2 (can use EEG system for this).
* [x] Modify RZ6.
* [x] Check RZ6 triggers still work.
* [ ] Check TDT as keyboard controller on MEG computer works (plugged into A1-A4, probably)

Closes #278.


Eric89GXL:
@rkmaddox do you have time to look at the new version of the circuit in RPvdsEx?

rkmaddox:
I will try very hard to look before I leave for my mega-trip.

Eric89GXL:
If not don't sweat it, we can debug in use too.

drammock:
having some trouble testing this.  `simple_experiment` run with a TDT reveals a bug with @ilogue's recent changes:

```
  File "C:\Users\labsner\Desktop\expyfun\examples\simple_experiment.py", line 82, in <module>
    pressed, timestamp = ec.wait_one_press(live_keys=live_keys)
  File "c:\users\labsner\python\expyfun\expyfun\_experiment_controller.py", line 1019, in wait_one_press
    relative_to)
  File "c:\users\labsner\python\expyfun\expyfun\_input_controllers.py", line 136, in wait_one_press
    pressed = self._correct_presses(pressed, timestamp, relative_to)[0]
  File "c:\users\labsner\python\expyfun\expyfun\_input_controllers.py", line 180, in _correct_presses
    events = [(k, s + self.time_correction, r) for k, s, r in events]
ValueError: need more than 2 values to unpack
```

drammock:
I think I've fixed the keyrelease errors, and have successfully tested the new circuit using `simple_experiment`.  Working from MARLO I'm unable to push to my own repo, so I pushed to a new branch on your remote @Eric89GXL.  It's called `test-circuit-buffer` and should be merged into this `buf` branch before I merge this into master.

Eric89GXL:
Go ahead and push directly to my existing branch if you're still at it, or
I'll do it in an hour or so

Eric89GXL:
Oh I just saw the PR, never mind

examples/simple_experiment.py
```diff
@@ -110,7 +110,7 @@
         ec.identify_trial(ec_id=stim_num, ttl_id=[0, 0])
         ec.write_data_line('one-tone trial', stim_num + 1)
         ec.start_stimulus()
-        pressed, timestamp = ec.wait_one_press(max_resp_time, min_resp_time,
+        pressed, timestamp, _ = ec.wait_one_press(max_resp_time, min_resp_time,
                                                live_keys)
```
drammock: oops, didn't indent.

Eric89GXL: got it

Eric89GXL:
@drammock if it ran okay on the RP2, I think the next step might be to check the RP2 triggers. This can be done e.g. by starting an experiment with the EEG powered on -- you should be able to see the triggers.

Assuming that works, you can do a behavioral pilot, and I can work on modifying the RZ6 version of the circuit. Sound good?

drammock:
Yep.

drammock:
I tested the JSALT experiment with the new circuit; triggers are definitely getting sent to the TDT and from there to the EEG.  I didn't yet check that they were in fact the correct bits, but will do so shortly by comparing to an actual subject session with the same stimuli, that ran on the old circuit.  In the meantime I noticed a problem: playing a short stim immediately after a longer one, I'm hearing the "ghost" of the longer one when the shorter one stops?  At least I think that's what's happening.  Definitely not obvious that auto-stop is auto-stopping at the right time, unless it's something special about the JSALT run script that is not playing friendly with the new circuit...  any ideas?

another problem is that running on the most recent commit of this PR (48475a8) the initial `screen_prompt` does an immediate flip so that the instructions are gone as soon as they appear (this doesn't happen on 0ee0951). I'll look into this and create a separate issue if merited (may be a

drammock:
@Eric89GXL do you have a test script you used with the RM1 to put this new circuit through its paces?

drammock:
confirmed using JSALT experiment script that triggers from new circuit are identical to triggers using old circuit.

Eric89GXL:
Sweet.

I did some testing of the circuit by putting in ones and zeros mostly, and ensuring that pops happened when they should. I don't think I have the script anymore, so I'll recreate it.

drammock:
I recommend testing with tones rather than clicks, e.g., load a 1 sec 1kHz tone, followed by a 0.5 sec 700 Hz tone, and see if you hear a tiny bit of the higher tone at the end of the (shorter) lower tone.

Eric89GXL:
@drammock it sounds okay on my RM1, can you check?
```
with ExperimentController('testExp', participant='foo', session='001',
                          output_dir=None, version='dev') as ec:
    ec.set_visible(False)
    t = np.arange(24414) / 24414.
    x = np.sin(2 * np.pi * 1000. * t)
    x[:12212] = np.sin(2 * np.pi * 700 * t[:12212])
    ec.load_buffer(0.01 * x)
    ec.play()
    ec.wait_secs(1.0)
    ec.stop()

    ec.load_buffer(0.01 * x[:12212])
    ec.play()
    ec.wait_secs(1.0)
    ec.stop()
```

drammock:
yeah, that sounds fine on MARLO.  Not sure what the problem was with the JSALT stims...  that exp. does some tricky things so I'll assume it was something weird with the timing of `wait_secs` and `stop` (still, puzzling that it doesn't happen on the older version).

drammock:
So, I'm noticing something weird with the new circuit: `ec.stop()` doesn't actually do anything to the playback right?  That's intentional?

```
import numpy as np
import expyfun.ExperimentController

with ExperimentController('testExp', participant='foo', session='001',
                          output_dir=None, version='dev') as ec:
    ec.set_visible(False)
    t = np.arange(24414) / 24414.
    x = np.sin(2 * np.pi * 1000. * t)
    x[:12212] = np.sin(2 * np.pi * 700 * t[:12212])
    ec.load_buffer(0.01 * x)
    ec.play()
    ec.wait_secs(0.01)
    ec.stop()  # this doesn't stop playback
    ec.wait_secs(1.5)

    ec.load_buffer(0.01 * x[:12212])
    ec.play()
    ec.wait_secs(1.5)
    ec.stop()
```

Eric89GXL:
No it should stop it, that's a bug. I'll fix it.

drammock:
that explains the weird JSALT behavior too.  I'm relieved.

Eric89GXL:
Okay @drammock the code snippet works now.

Eric89GXL:
I also got rid of the `pause/reset` in `_tdt_controller`, now anytime you stop it resets. It simplifies our code a little bit, and I don't think we currently have use cases for "pausing" playback.

drammock:
so just so that I understand the new system: if you load `N` samples into the buffer, the circuit will auto-stop after the `N`th sample is played,  and return the "playhead" to zero.  If you want to, you can call `ec.stop` earlier than the `N`th sample, which will send a trigger to the circuit causing playback to stop (and also return the playhead to zero).  But if you don't do that, the circuit will auto-stop and auto-reset, but `ec._playing` will still be `True` so you'll need to (redundantly) call `ec.stop` anyway?

expyfun/_tdt_controller.py
```diff
@@ -212,12 +212,7 @@ def play(self):
...
-        """Stop playback and reset the buffer position"""
-        self.pause()
-        self.reset()
-
-    def pause(self):
-        """Send the soft trigger to stop the ring buffer playback.
+        """Send the soft trigger to stop and reset the ring buffer playback.
         """
         self._trigger(2)
         logger.debug('Stopping TDT audio')
```
drammock: for consistency this should say `'Expyfun: Stopping TDT audio'`

Eric89GXL:
Yeah `ec._playing` will be misleading. We could "fix" it by polling the TDT circuit, but I don't like that too much. I don't mind forcing people to do a redundant `ec.stop` call because it's more explicit.

Eric89GXL:
(Note that `ec._playing' is similarly misleading for the Pyglet audio backend)

Eric89GXL:
@drammock debug message fixed. Let me know if the TDT is happy with it. Then I'll work on the RZ6 circuit, and maybe come up with some test script that you can run to validate it.

BTW, should we also add a pre-experiment sanity check script here? It's something I've been meaning to add for a while, e.g. do a tone pip on the left and right, experimenter manually advances when the subject responds properly. We could add the blink instructions, too, if we want, but we at least should make sure the left/right is correct.

drammock:
I like the idea of a codeblock that could be used for verifying that L/R audio are both working.  An additional MEG-specific codeblock with textual prompts for pushing the buttons, moving dots for looking left/right, and tones for blinking might work well.

Regarding the autostop issue: is there any reason at all to keep `ec._playing` (and the check against it within `ec._play()`)?  People already have to keep track of audio durations via `ec.wait_secs` or `ec.wait_until` for timing purposes, so I don't see the advantage of forcing an `ec.stop` to be in there every time.

drammock:
confirmed that stopping works properly now on MARLO / RP2.1

Eric89GXL:
Yeah, it might be worth trying to get rid of the absolute requirement of `.stop`. But we probably don't want people doing `.load_buffer` before the stimulus is done playing, and we probably don't want them doing that before `.trial_ok()` is called -- and we need `ec._playing` for that. I also don't know for sure if Pyglet exposes whether or not a stimulus completes, but I'm pretty sure we can figure that out.

@rkmaddox WDYT, you okay with the auto-stop so people don't have to have `ec.stop()` in their code anymore?

drammock:
thinking about the ONR timeline, maybe this deserves a separate issue, to
be fixed later.  Sorry for the tangent (though still want to hear what
@rkmaddox thinks)

Eric89GXL:
BTW I talked to @mfreilly today and he's going to look into teeing the MEG button box to the TDT as well as the Neuromag system so we don't have to rely on the box-as-keyboard as an input device anymore. Plus we'll actually have reliable timing for button presses. I'm going to translate our existing button-press code for RP2 over to the RZ6 circuit while I'm in there making the auto-stop changes.

rkmaddox:
That's tricky. We still want people to be able to issue a stop command.
Maybe when playing on tdt expyfun should continuously poll the circuit to
see if it's stopped? Or we could have it automatically go into a stop state
after the appropriate amount of time, without sending a stop message to the
circuit. If implemented carefully, that would result in less than a couple
ms of difference.
On Apr 27, 2016 4:56 PM, "Eric Larson" <notifications@github.com> wrote:

BTW I talked to @mfreilly <https://github.com/mfreilly> today and he's
going to look into teeing the MEG button box to the TDT as well as the
Neuromag system so we don't have to rely on the box-as-keyboard as an input
device anymore. Plus we'll actually have reliable timing for button
presses. I'm going to translate our existing button-press code for RP2 over
to the RZ6 circuit while I'm in there making the auto-stop changes.

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/LABSN/expyfun/pull/266#issuecomment-215267400>

Eric89GXL:
You can still manually stop early under the proposed changes, but it would
no longer be mandatory to call stop every trial.

rkmaddox:
That is ideal. I was reading on my phone, wasn't the question what to do
about the `_playing` status? If we want to keep that status we could have
expyun issue its stop automatically as well. Sorry, I may just be behind on
the conversation.

Eric89GXL:
Okay @drammock RZ6 circuit updated. Turns out the button press stuff was already in there. You might be able to try plugging the button box into the RZ6 input row "A" (first four probably?) to see if e.g. the keypress demo works. If it does, then the "tee" solution Myles is working on should allow us to change the config on that machine to use the TDT as a keypress controller (hooray!). So add that to the list of tests to run, in addition to the simple AV-sync and trigger tests.

Should we augment https://github.com/LABSN/expyfun/blob/master/examples/experiments/sync_test.py to have triggers 2, 4, and 8 stamped in there somewhere?

mfreilly:
Just tried plugging in to A1 - A4 and running key press demo. It waits for
key presses but does not see any. However one of the 4-button boxes does
enter text numbers into the script editor.
-- 
Myles Reilly
Research Scientist
UW I-LABS
MEG Brain Imaging Center
Portage Bay Bldg 244
206-685-3395



From:  Eric Larson <notifications@github.com>
Reply-To:  LABSN/expyfun
<reply+00aaa70770cd97b31e52463e2d2e883177934a562c1662c192cf00000001133a26ba9
2a169ce07fe37c1@reply.github.com>
Date:  Thursday, April 28, 2016 at 12:30 PM
To:  LABSN/expyfun <expyfun@noreply.github.com>
Cc:  Myles Reilly <mreilly2@uw.edu>, Mention <mention@noreply.github.com>
Subject:  Re: [LABSN/expyfun] FIX: Bigger buffer and auto-stop (#266)


Okay @drammock <https://github.com/drammock>  RZ6 circuit updated. Turns out
the button press stuff was already in there. You might be able to try
plugging the button box into the RZ6 input row "A" (first four probably?) to
see if e.g. the keypress demo works. If it does, then the "tee" solution
Myles is working on should allow us to change the config on that machine to
use the TDT as a keypress controller (hooray!). So add that to the list of
tests to run, in addition to the simple AV-sync and trigger tests.

Should we augment 
https://github.com/LABSN/expyfun/blob/master/examples/experiments/sync_test.
py to have triggers 2, 4, and 8 stamped in there somewhere?


You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/LABSN/expyfun/pull/266#issuecomment-215536898>

Eric89GXL:
@mfreilly you'll have to test it with the latest changes from this branch, did you do that? @drammock will probably come by in the coming days to do it, because if it doesn't work, it will require some debugging.

Eric89GXL:
@drammock requirement for `ec.stop` removed, and `sync_test.py` has been augmented with `ec.stamp_triggers([2, 4, 8])` so hopefully that facilitates your testing.

drammock:
@Eric89GXL triggers look normal during synctest:

triggers

However, the MEG button boxes are not functioning; buttons 2 & 3 (yellow, green) appear to be stuck in "on" position on one controller, and button 1 (red) is stuck on the other one (not mechanically stuck --- they're not getting through to the response box, and when its output is plugged directly into TDT BNC ports A1-A4, the stuck buttons send a constant "on" signal). Even when the response box is connected to NINTOK via USB, we still see the same pattern of buttons 2 & 3 not working (not being registered by expyfun).  @mfreilly is looking into this.

Moreover, even the buttons that work (i.e., that light up the appropriate input LED on the RZ6) are failing to cause the tones to play during the training phase of `simple_experiment`.  Not sure if that will resolve itself once the buttons are fixed, or if somehow the circuit is failing to pass the input triggers to expyfun.

Eric89GXL:
Just to double-check -- did you run with `response_device='tdt'`?

If so, the next check might be to run the circuit in RPSvdEX on the MEG machine, and add a monitor line to look at the inputs in realtime. Have you worked with the TDT stuff at all before?

drammock:
umm... no.  I'm so used to it being set in the config file, and it was
clear that the audio was coming out of the TDT based on other tests I ran;
I guess I assumed it would default to using TDT as response device once we
had unplugged the USB connection to the response box.  I'll try again later
today.

mfreilly:
I got one response pad working on all 4 buttons (8765). The config on the
response box had to be reset.
It is sending triggers to the TDT. The other box still only has 2 working
buttons, not sure why.
-- 
Myles Reilly
Research Scientist
UW I-LABS
MEG Brain Imaging Center
Portage Bay Bldg 244
206-685-3395



From:  Eric Larson <notifications@github.com>
Reply-To:  LABSN/expyfun
<reply+00aaa707bd52acaeb05e249c5d387261dd06d427da6a95d592cf0000000113409fda9
2a169ce07fe37c1@reply.github.com>
Date:  Tuesday, May 3, 2016 at 10:20 AM
To:  LABSN/expyfun <expyfun@noreply.github.com>
Cc:  Myles Reilly <mreilly2@uw.edu>, Mention <mention@noreply.github.com>
Subject:  Re: [LABSN/expyfun] FIX: Bigger buffer and auto-stop (#266)


Just to double-check -- did you run with response_device='tdt'?

If so, the next check might be to run the circuit in RPSvdEX on the MEG
machine, and add a monitor line to look at the inputs in realtime. Have you
worked with the TDT stuff at all before?


You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/LABSN/expyfun/pull/266#issuecomment-216601807>

drammock:
@Eric89GXL there are still at least two problems.

1. Although the response box is receiving all button presses (as indicated by its front-panel LEDs) they are not reliably being passed on to the TDT; long hard presses seem to always get through, but soft or fast presses sometimes light up the response box but do not light the corresponding input bit on the RZ6.  Maybe the response box output voltage is on the weak side?  A [response box upgrade](http://www.curdes.com/mainforp/interfaces/932-upgrade.html) may be worth investigating.

2. Even when the presses make it through to the TDT (as indicated by the input bit LEDs 0-3 lighting up) they are not registered by expyfun (even with `response_box='tdt'`): presses are not appearing in the log or output files, and are not triggering stimulus playback during `simple_experiment` training.

Eric89GXL:
> Maybe the response box output voltage is on the weak side? A response box upgrade may be worth investigating.

It's $200 which is fairly inexpensive. From the description on the website, it sounds like a worthwhile upgrade even if it doesn't fix our problem. @mfreilly do you agree?

In the meantime, we could see if there is a way to make the Neuromag system pump a trigger out from one of its output ports any time it receives an input. Maybe it does this already (duplicates input to an output port) somehow for chaining. @mfreilly any ideas there?

> Even when the presses make it through to the TDT (as indicated by the input bit LEDs 0-3 lighting up) they are not registered by expyfun (even with response_box='tdt'): presses are not appearing in the log or output files, and are not triggering stimulus playback during simple_experiment training.

@drammock unfortunately I don't think I can debug this remotely. Either you or @rkmaddox will need to fire up RPVDEX, load and run the circuit there, and look at the input bits in real-time (by adding a monitor tag) to see where the problem is. It might require reconfiguring the circuit to receive them properly. Have you tried using RPVDEX before? It's a bit annoying but not too complicated overall.

mfreilly:
There are output TTL signals from the Neuromag trigger box, these can be
connected directly to the TDT inputs.
I¹ll look at upgrade.
-- 
Myles Reilly
Research Scientist
UW I-LABS
MEG Brain Imaging Center
Portage Bay Bldg 244
206-685-3395



From:  Eric Larson <notifications@github.com>
Reply-To:  LABSN/expyfun
<reply+00aaa7079ad0b5eae5edc8823744288291640a1efd38706592cf000000011341f32d9
2a169ce07fe37c1@reply.github.com>
Date:  Wednesday, May 4, 2016 at 10:28 AM
To:  LABSN/expyfun <expyfun@noreply.github.com>
Cc:  Myles Reilly <mreilly2@uw.edu>, Mention <mention@noreply.github.com>
Subject:  Re: [LABSN/expyfun] FIX: Bigger buffer and auto-stop (#266)

> 
> Maybe the response box output voltage is on the weak side? A response box
> upgrade may be worth investigating.

It's $200 which is fairly inexpensive. From the description on the website,
it sounds like a worthwhile upgrade even if it doesn't fix our problem.
@mfreilly <https://github.com/mfreilly>  do you agree?

In the meantime, we could see if there is a way to make the Neuromag system
pump a trigger out from one of its output ports any time it receives an
input. Maybe it does this already (duplicates input to an output port)
somehow for chaining. @mfreilly <https://github.com/mfreilly>  any ideas
there?
> 
> Even when the presses make it through to the TDT (as indicated by the input
> bit LEDs 0-3 lighting up) they are not registered by expyfun (even with
> response_box='tdt'): presses are not appearing in the log or output files, and
> are not triggering stimulus playback during simple_experiment training.

@drammock <https://github.com/drammock>  unfortunately I don't think I can
debug this remotely. Either you or @rkmaddox <https://github.com/rkmaddox>
will need to fire up RPVDEX, load and run the circuit there, and look at the
input bits in real-time (by adding a monitor tag) to see where the problem
is. It might require reconfiguring the circuit to receive them properly.
Have you tried using RPVDEX before? It's a bit annoying but not too
complicated overall.


You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/LABSN/expyfun/pull/266#issuecomment-216940209>

drammock:
> Have you tried using RPVDEX before?
Yes, in fact I looked at the circuit this morning on NINTOK and was stymied
by how little I remember about how it all works.  It will take me a while
to get back up to speed on RPVDEX, definitely won't get to it this week...

Eric89GXL:
@rkmaddox is a TDT circuit whiz you should bother him :)

rkmaddox:
Yeah, we can take a look at it together. I haven't looked at it in ages,
but have spent enough time staring in baffled bemusement that I have a
baseline memory of it.

--
Ross Maddox, Ph.D.
Postdoctoral Fellow
University of Washington
phone: 206-685-4662

Eric89GXL:
@drammock where are we on this one? Can we merge and sort out the TDT as keyboard stuff later?

Eric89GXL:
@mirbagheri are you using this branch for your experiment?

mfreilly:
The monster lives! Everything reinstalled sounds and triggers work.
Thanks for your time Eric, have a good weekend 😊

drammock:
@Eric89GXL  I guess I'm OK with merging this as auto-stop.  I'll open a new issue for tying button-box to TDT for response times.

Eric89GXL:
Thanks for putting in the testing work on this PR @drammock, and for helping @mfreilly work things out today

PR #268, opened 2016-02-26
WIP: Add video

### LINKED ISSUE #267: adding video support

    - `ec.load_movie()`
    - etc

Thoughts?

drammock:
WIP do not merge. Have a look if you want. Docs need work; no seeking or scaling (yet) and some issues with passing control back to the user.

Closes #267


Eric89GXL:
Does it actually display now?

drammock:
yes!  definitely some debugging to be done w/ different video encodings...  the example video appears to have a transparent background (which looks weird) but the ones I'll be using in the experiment are working well so far.

examples/stimuli/video_playback.py
```diff
@@ -0,0 +1,29 @@
...
+This script does XXX.
+"""
+# @author: drmccloy
+# Created on Tue Feb 23 13:57:53 2016
+# License: BSD (3-clause)
+
+from __future__ import print_function
+from expyfun import ExperimentController as EC
+
+movie_path = './example-movie.mp4'
```
Eric89GXL: How big is this movie? Could we download one on-the-fly from the internet to avoid repo bloat?

examples/stimuli/video_playback.py
```diff
@@ -0,0 +1,29 @@
...
+               full_screen=False, participant='foo', session='foo',
+               version='dev', enable_video=True)
+ec = EC(**ec_args)
+ec.load_video(movie_path)
+ec.video.play()
+#ec.video.seek(15.)
+
+while ec.video.playing:
+    next_frame = ec.video.last_frame + ec.video.dt
+    _ = ec.flip(when=next_frame)
```
Eric89GXL: if you don't want the output you can just do `ec.flip(...)` with no `_ =` preceding it.

expyfun/_experiment_controller.py
```diff
@@ -840,13 +852,14 @@ def flip(self, when=None):
...
-        gl.glEnd()
-        gl.glFinish()
+        if not self._enable_video:
+            # this waits until everything is called, including last draw
+            gl.glClear(gl.GL_COLOR_BUFFER_BIT)
+            gl.glBegin(gl.GL_POINTS)
+            gl.glColor4f(0, 0, 0, 0)
+            gl.glVertex2i(10, 10)
+            gl.glEnd()
+            gl.glFinish()
```
Eric89GXL: Did you need to uncomment all lines? Maybe `glEnd` and `glFinish` can stay unindented?

expyfun/_experiment_controller.py
```diff
@@ -840,13 +852,14 @@ def flip(self, when=None):
...
-        # this waits until everything is called, including last draw
-        gl.glClear(gl.GL_COLOR_BUFFER_BIT)
-        gl.glBegin(gl.GL_POINTS)
-        gl.glColor4f(0, 0, 0, 0)
-        gl.glVertex2i(10, 10)
-        gl.glEnd()
-        gl.glFinish()
+        if not self._enable_video:
+            # this waits until everything is called, including last draw
+            gl.glClear(gl.GL_COLOR_BUFFER_BIT)
```
Eric89GXL: maybe this is the only one you need to indent?

Eric89GXL:
Looks like a reasonable start to me

examples/stimuli/video_playback.py
drammock: yeah, was planning to switch it to be handled like the HRTFs (DL from lester) before merging into master.

Eric89GXL: Okay don't forget to squash commits to save history

PR #270, opened 2016-03-01
Add video (squashed)

### LINKED ISSUE #267: adding video support

drammock:
Closes #268. Closes #267.

Ready for review @Eric89GXL @rkmaddox.  Please be sure to `make test` locally; pyglet is something of a moving target so I don't want any surprises due to different pyglet versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ before_install:
     # install AVbin
     - wget https://github.com/downloads/AVbin/AVbin/install-avbin-linux-x86-64-v10 -O avbin.sh
     - chmod +x avbin.sh
-    - sudo ./avbin.sh
+    - yes | sudo ./avbin.sh
     # Only install these deps if necessary (slow on Py3k)
     - travis_retry pip install -q nose-timer flake8;
     - travis_retry pip install -q pyglet mne six;  # mne is a dependency now that resample is taken from it
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,10 @@ before_install:
         conda install --yes scipy matplotlib nose coverage setuptools;
         travis_retry pip install -q coveralls;
       fi;
+    # install AVbin
+    - wget https://github.com/downloads/AVbin/AVbin/install-avbin-linux-x86-64-v10 -0 avbin.sh
+      chmod +x avbin.sh
+      ./avbin.sh
     # Only install these deps if necessary (slow on Py3k)
     - travis_retry pip install -q nose-timer flake8;
     - travis_retry pip install -q pyglet mne six;  # mne is a dependency now that resample is taken from it
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
         travis_retry pip install -q coveralls;
       fi;
     # install AVbin
-    - wget https://github.com/downloads/AVbin/AVbin/install-avbin-linux-x86-64-v10 -O avbin.sh
+    - wget http://github.com/downloads/AVbin/AVbin/install-avbin-linux-x86-64-v10 -O avbin.sh
     - chmod +x avbin.sh
     - ./avbin.sh
     # Only install these deps if necessary (slow on Py3k)
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: required
+dist: trusty
 language: python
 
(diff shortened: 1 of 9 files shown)


drammock:
Travis is choking on the video test (`WAVEFormatException: AVbin is required to decode compressed media`).  @Eric89GXL you're the travis-master, how should I handle?  Installing AVbin is done via executable binary [here](http://avbin.github.io/AVbin/Download.html) (currently v10) or from linux repos as libavbin0 (currently version 7).  Should I test video only when `DEPS="full"`?

AppVeyor, on the other hand, is choking on `test_logging`?

Eric89GXL:
I'd try the executable version it's more likely to work.

I'll try to take a look at the logging test soon, just ignore it for now

.travis.yml
```diff
@@ -32,6 +32,10 @@ before_install:
         conda install --yes scipy matplotlib nose coverage setuptools;
         travis_retry pip install -q coveralls;
       fi;
+    # install AVbin
+    - wget https://github.com/downloads/AVbin/AVbin/install-avbin-linux-x86-64-v10 -O avbin.sh
```
drammock: @Eric89GXL any idea why this `wget` command returns "4" (network error) all 3 times that travis tried it, but works fine locally?

Eric89GXL: The error is `Unable to establish SSL connection`, which suggests that it's an SSL protocol failure. Either a bug in Ubuntu/wget or GitHub's certs. Can you use http?

drammock: Thanks, I'll look into that. How did you figure out it was ssh? All I saw
in the logs was "error code 4" which is supposedly a generic network error.
How could I have dug deeper?

Eric89GXL:
I looked here:

https://travis-ci.org/LABSN/expyfun/jobs/112970700#L279

What logs did you look at?

drammock:
Same logs. Apparently I was blind to all text that was not highlighted red.
Sorry.

rkmaddox:
Tested on my machine. No errors in `noestests`. One warning on every play:

```python
/home/rkmaddox/anaconda2/lib/python2.7/site-packages/pyglet/media/__init__.py:1170: DeprecationWarning: Player.eos_action is deprecated in favor of SourceGroup.loop and SourceGroup.advance_after_eos
  category=DeprecationWarning)
```

Playback was choppy on my system, compared to playback in mplayer, so it appears it was missing several frames. The good news is that playback length was unaffected, so it seems to sticking to master clock, rather than just playing the next frame in the sequence regardless of when the previous frame was.

drammock:
thanks for pointing out the choppiness, @rkmaddox.  I'll try running on
MARLO to see if this is just the "linux halves the frame rate" issue, or
something more insidious.

rkmaddox:
The choppiness did seem regular, so it could be a half (or quarter?) framerate problem.

drammock:
we'll see if the CIs are happy now.  @rkmaddox that warning you mentioned is coming from `pyglet.__init__` which apparently didn't get the memo about `eos_action` being deprecated (FWIW, it looks like the replacement is not really fully implemented yet either, so for now we have to survive without end-of-stream events getting issued).

drammock:
After some digging: the travis failures are probably due to an old bug in `wget`.  We're running the tests on 12.04 (precise) which ships with wget version 1.13.4-2ubuntu1.2.  Apparently 14.04 (trusty) ships with a version of `wget` that has the bug fixed (1.15-1ubuntu1.14.04.1).  So options are:

1. run our tests on trusty instead of precise
1.  pass `--no-check-certificate` to `wget`

@Eric89GXL do you have a preference between those two options?  Note that `apt-get install wget` will not work, since no newer version of `wget` was ever released in precise-updates or precise-backports.

Eric89GXL:
I didn't know Travis had an OS version option. Does it?

I'm fine with the no check option

drammock:
you can specify it in `.travis.yml`:
https://docs.travis-ci.com/user/trusty-ci-environment/

Eric89GXL:
Didn't know that -- I wonder when they rolled it out. Would have saved me a
lot of pain some months ago having to deal with old 12.04 junk...

Anyway, +1 to try the upgrade.

drammock:
here goes nothin'!

PR #272, opened 2016-03-04
MRG: New video implementation

### LINKED ISSUE #267: adding video support

### LINKED ISSUE #269: can't easily remove on_every_flip functions
I want to add an element to `ec.on_every_flip_functions`, then later remove it.  I don't think there's a good way to do that, since it's a `@property`.  In theory I can access private list `ec._on_every_flip`, search through it, and `pop` the appropriate element (which for the current use case is fine because I'm doing this within `_visual.py`).  That *would* work well if I called `functools.partial` *before* passing into `ec.call_on_every_flip` so that I could compare the objects directly, but it's less straightforward because `ec.call_on_every_flip` calls `functools.partial` internally, making it a bit ugly to keep track of what gets passed in, so as to later compare `ec._on_every_flip` list elements to it. Example:

```python
# here self._draw is <type: functools.partial> already
for idx, func in enumerate(self._ec._on_every_flip):
    if hasattr(func, 'func'):
        if func.func is self._draw:
            self._ec._on_every_flip.pop(idx)
```

possible changes/solutions:

1. make `ec.on_every_flip_functions` read/write, possibly by adding a new `remove_on_every_flip` method.
1. alter `ec.call_on_every_flip` to return the partialized function, so you can compare against it with `is` later when iterating through `ec._on_every_flip`.
1. alter `ec.call_on_every_flip` to generate a dict instead of list, and have it return a key, so it can be deleted easily later.
1. alter `ec.call_on_every_flip` to check whether its incoming function is already a partial, and if so, don't call partial on it again.
1. alter `ec.call_on_every_flip` to not call partial at all, and leave it up to the user to call it before passing in the function.

Thoughts?

drammock:
This PR is a (hopefully) improved implementation of pyglet video playback.  The PRs for the other two attempts have been closed without merging.  It also changes the behavior of `ec.call_on_next_flip` and `ec.call_on_every_flip` by removing the internal call to `functools.partial` (see #269); henceforth, users must call `partial` (as needed) before adding functions to the on_next / on_every queue.  This has the advantage of allowing functions in the queue to be easily detected (via `ec.on_every_flip_functions.index()` and subsequently `pop()`ped.

The attempts at running travis with `trusty` caused pip to choke on mne, so I'm reverting to `precise` and avoiding the wget error via `--no-check-certificate`.

Closes #267 
Closes #269

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ before_install:
     # install AVbin
     - curl https://github.com/downloads/AVbin/AVbin/install-avbin-linux-x86-64-v10 -o avbin.sh
     - chmod +x avbin.sh
-    - yes | sudo ./avbin.sh
+    - sudo sh -c "yes | ./avbin.sh"
     # Only install these deps if necessary (slow on Py3k)
     - travis_retry pip install -q nose-timer flake8;
     - travis_retry pip install -q pyglet mne six;  # mne is a dependency now that resample is taken from it
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ before_install:
     # install AVbin
     - curl https://github.com/downloads/AVbin/AVbin/install-avbin-linux-x86-64-v10 -o avbin.sh
     - chmod +x avbin.sh
-    - sudo sh -c "yes | ./avbin.sh"
+    - sudo bash -c "yes | ./avbin.sh"
     # Only install these deps if necessary (slow on Py3k)
     - travis_retry pip install -q nose-timer flake8;
     - travis_retry pip install -q pyglet mne six;  # mne is a dependency now that resample is taken from it
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,15 @@ before_install:
         conda install --yes scipy matplotlib nose coverage setuptools;
         travis_retry pip install -q coveralls;
       fi;
+    # install a newer version of wget (avoid SSL errors)
+    - mkdir wget
+    - wget http://ftp.gnu.org/gnu/wget/wget-1.17.tar.xz -O wget.xz
+    - tar xf wget.xz -C ./wget --strip-components=1
+    - cd wget && ./configure && make
+    - sudo make install
+    - cd
     # install AVbin
-    - wget --no-check-certificate https://github.com/downloads/AVbin/AVbin/install-avbin-linux-x86-64-v10 -O avbin.sh
+    - wget https://github.com/downloads/AVbin/AVbin/install-avbin-linux-x86-64-v10 -O avbin.sh
     - chmod +x avbin.sh
     - yes | sudo ./avbin.sh
     # Only install these deps if necessary (slow on Py3k)
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,9 @@ before_install:
     - wget https://github.com/downloads/AVbin/AVbin/install-avbin-linux-x86-64-v10 -O avbin.sh
     - chmod +x avbin.sh
     - yes | sudo ./avbin.sh
+    # fix for InsecurePlatformWarning in pip
+    - travis_retry pip install pyopenssl pyasn1
+    - travis_retry pip install --upgrade ndg-httpsclient
     # Only install these deps if necessary (slow on Py3k)
     - travis_retry pip install -q nose-timer flake8;
     - travis_retry pip install -q pyglet mne six;  # mne is a dependency now that resample is taken from it
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,9 @@ before_install:
         travis_retry pip install -q coveralls;
       fi;
     # install AVbin
-    - curl https://github.com/downloads/AVbin/AVbin/install-avbin-linux-x86-64-v10 -o avbin.sh
+    - curl https://cloud.github.com/downloads/AVbin/AVbin/install-avbin-linux-x86-64-v10 -o avbin.sh
     - chmod +x avbin.sh
-    - sudo bash -c "yes | ./avbin.sh"
+    - yes | sudo tee /dev/null | ./avbin.sh
     # Only install these deps if necessary (slow on Py3k)
     - travis_retry pip install -q nose-timer flake8;
     - travis_retry pip install -q pyglet mne six;  # mne is a dependency now that resample is taken from it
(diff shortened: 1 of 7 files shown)


drammock:
OK, I'm getting rather frustrated with Travis.  here is the current state of affairs regarding AVbin:

- vanilla `wget` fails SSH handshake
- adding `--no-check-certificate` [doesn't help](https://travis-ci.org/LABSN/expyfun/jobs/113557127#L270)
- downloading and building a more recent `wget` into the Travis container failed, apparently because [wget expects/requires py3k to be the default python during its build](https://travis-ci.org/LABSN/expyfun/jobs/113714771#L336)?  I'm surprised python is even needed to build wget, but whatever.
- running Travis with trusty instead of precise [fixes the wget issues](https://travis-ci.org/LABSN/expyfun/jobs/113469792#L372), but [causes SSL problems (InsecurePlatformWarning) with pip](https://travis-ci.org/LABSN/expyfun/jobs/113469792#L573).  Those problems are [allegedly fixed in python 2.7.9](https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning) and later, but (and this is really odd) [the Travis blog claims](https://docs.travis-ci.com/user/trusty-ci-environment/#Python) that Travis-trusty containers use python 2.7.10, but [the build log](https://travis-ci.org/LABSN/expyfun/jobs/113469792#L165) says it's running 2.7.6.

WTF?

drammock:
Meanwhile, AppVeyor continues its streak of randomly failing every 3-4 builds on different parts of the test suite each time.

Eric89GXL:
I remember having problems with `wget` and getting around them using `curl`

drammock:
wget errors are fixed now.  Fingers crossed on this latest commit...

drammock:
woohooo!  in your face Travis!  @Eric89GXL @rkmaddox this is FINALLY ready for review.

.travis.yml
```diff
@@ -32,6 +32,10 @@ before_install:
         conda install --yes scipy matplotlib nose coverage setuptools;
         travis_retry pip install -q coveralls;
       fi;
+    # install AVbin
+    - curl https://cloud.github.com/downloads/AVbin/AVbin/install-avbin-linux-x86-64-v10 -o avbin.sh
+    - chmod +x avbin.sh
+    - sudo bash -c "yes | ./avbin.sh"
```
Eric89GXL: This is awesome

expyfun/visual/_visual.py
```diff
@@ -935,3 +931,216 @@ def draw(self):
...
+    def pause(self):
+        """Halt video playback."""
+        if self._playing:
+            idx = self._ec.on_every_flip_functions.index(self.draw)
+            self._ec.on_every_flip_functions.pop(idx)
+            self._player.pause()
+            self._playing = False
+        else:
+            raise RuntimeWarning('ExperimentController.video.pause() called '
+                                 'when already paused.')
```
Eric89GXL: This should probably be warnings.warn()

expyfun/visual/_visual.py
```diff
@@ -935,3 +931,216 @@ def draw(self):
...
+        if self._texture is not None:
+            self._texture.width = self.source_width * self._scale
+            self._texture.height = self.source_height * self._scale
+
+    def set_scale(self, scale=1., fill=False):
+        """Set video scale.
+
+        Parameters
+        ----------
+        scale : float
```
Eric89GXL: How about scale can be float or str ("fill")?

Eric89GXL:
Otherwise LGTM. Let me know when those comments are addressed and I'll pull test and merge

.travis.yml
drammock: thanks.  FYI, using `curl` instead of `wget` had hiccups of its own: the public URL (`github.com/downloads/...`) redirected to the `cloud.github.com/...` address seen here, and `curl` was initially downloading only the redirect page, not its target (`wget` seems to have been clever enough to follow the redirect).  Noting here, as I don't know how stable the `cloud.github.com` address will be in future.

expyfun/_experiment_controller.py
```diff
@@ -843,7 +852,8 @@ def flip(self, when=None):
         # this waits until everything is called, including last draw
         gl.glClear(gl.GL_COLOR_BUFFER_BIT)
         gl.glBegin(gl.GL_POINTS)
-        gl.glColor4f(0, 0, 0, 0)
+        if not self._enable_video:
+            gl.glColor4f(0, 0, 0, 0)
```
drammock: I noticed that changing the value of `_enable_video` has no effect once `ExperimentController` has been instantiated (presumably because `ec.flip()` will only get evaluated once, at init time, at which point the answer to `if not self._enable_video` is baked in).  I haven't fully tested the ramifications of omitting the `gl.glColor4f(0, 0, 0, 0)` for non-video stimuli, but I wonder if it's worth omitting the `enable_video` param and having separate `flip()` and `flip_video()` methods?  Seems to violate DRY, but might allow more flexible experiment designs.  YAGNI?

Eric89GXL: That would have a nice side benefit of making it clear that you shouldn't combine draws and video display

drammock: I decided not to change this right now, in the interest of getting on with my experiment.  Will open a TODO issue for testing video / drawing / audio synchronization, and implement separation this at that point if needed.  Check the new example though: drawing text on top of video seems to work fine, so I'm not convinced that the separation is necessary.

drammock:
arrrgh Travis!  py2k builds stalled on a line in `.travis.yml` that I didn't touch.  Already tried restarting those jobs once, just tried a second time (hoping it's just an ubuntu or pypi server being down or something).  If not, may need your advice @Eric89GXL

drammock:
OK @Eric89GXL, the Travis stalls are occurring *before* it gets to the installation of AVbin, and the logs are not telling me much other than "I waited 10 minutes and nothing happened, terminating as stalled build."  help?  I've restarted those builds twice already and got the same result all 3 times.

Eric89GXL:
I restarted the builds and it's fine. I guess Travis likes some people better than others.

Good to go from your end?

drammock:
Yes!

Eric89GXL:
Thanks @drammock

PR #275, opened 2016-03-15
MRG: show/hide video

drammock:
adds show/hide functionality to videos, and adds compatibility between video playback and screen background colors.  Also splits the video playback example into "simple" and "advanced" variants.


drammock:
@Eric89GXL can you grant me access to the AppVeyor expyfun project, so I can restart AppVeyor builds when they fail for random reasons?  Other than that I think this PR is ready for review/merge.  @rkmaddox if you have time would like your input on the video API decisions implicit in this PR.

Eric89GXL:
That's weird, you should be able to since you're an admin. For now you can do the dirty trick of `git commit --amend -am '<Old commit comment'` then `git push origin --force branch-name`.

expyfun/visual/_visual.py
```diff
@@ -946,16 +946,21 @@ class Video(object):
...
-    scale : float
+    scale : float | str
         The scale factor. 1 is native size (pixel-to-pixel), 2 is twice as
-        large, etc. Ignored if ``fill_window`` is ``True``.
+        large, etc. If `scale` is a string, it must be either ``'fill'``
+        (which ensures the entire ``ExperimentController`` window is
+        covered by the video, at the expense of some parts of the video
+        potentially being offscreen), or ``'fit'`` (which scales maximally
+        while ensuring none of the video is offscreen, which may result in
+        letterboxing).
```
Eric89GXL: letterboxing or pillarboxing :)

expyfun/visual/_visual.py
```diff
@@ -1086,14 +1116,25 @@ def draw(self):
...
+            if self._visible:
+                self._texture = self._player.get_texture()
+                self._scale_texture()
+                self._texture.blit(*self._actual_pos)
         else:
             self._finished = True
             self.pause()
         self._ec.check_force_quit()
 
+    def set_visible(self, show, flip=False):
```
Eric89GXL: if `flip=False` by default, is there much advantage to:
```
video.set_visible(True, flip=True)
```
over
```
video.set_visible(True)
ec.flip()
```
? I guess it saves a line, is that the idea?

Eric89GXL:
LGTM but please add some simple tests

expyfun/visual/_visual.py
drammock: Is pillarboxing even a real word?

expyfun/visual/_visual.py
drammock: The idea of having the flip arg is to remind users that it won't take effect until the next flip. The idea of defaulting to false is to not do things like flips automatically without very strong reasons. That was my thinking anyway, but I'm happy to defer if you think it is silly or confusing.

Eric89GXL: Nah, that makes sense as a reason

expyfun/visual/_visual.py
Eric89GXL: http://lmgtfy.com/?q=pillarboxing

expyfun/visual/_visual.py
```diff
@@ -991,25 +991,10 @@ def __init__(self, ec, file_name, pos=(0, 0), units='norm', scale=1.,
         self._units = units
         self._center = center
         self.set_scale(scale)  # also calls set_pos
-        self.set_visible(visible)
-        self._update_background()
-
-    def _update_background(self):
```
drammock: This turned out not to be working anyway, and the `GLubyte` operation is sloooow.  Hit all sorts of pyglet errors trying to just change the already existing texture's image data rather than creating a new `ImageData` object, and eventually gave up.

expyfun/visual/_visual.py
```diff
@@ -991,25 +991,10 @@ def __init__(self, ec, file_name, pos=(0, 0), units='norm', scale=1.,
...
-    def play(self, autoshow=True):
-        """Play video from current position.
+        self._visible = visible
 
-        Parameters
-        ----------
-        autoshow : bool
-            If ``True`` (and video is currently hidden), will make video
-            visible before playing.
+    def play(self):
```
drammock: `play(autoshow)` and `pause(autohide)` were causing more problems than they were worth.

expyfun/visual/_visual.py
```diff
@@ -1111,27 +1094,31 @@ def set_pos(self, pos, units='norm', center=True):
...
+        """Show/hide the video frame."""
         if show:
             self._visible = True
+            self._draw()
         else:
-            self._texture.blit_into(self._background, 0, 0, 0)
             self._visible = False
+            self._ec.flip()
         if flip:
             self._ec.flip()
```
drammock: this is super annoying, but for the life of me I can't get `set_visible(False)` to work without doing **two** flips (related to earlier comment about not being able to alter the `ImageData` in an existing pyglet `TextureRegion`).  Ideas for alternate approaches welcome.

examples/stimuli/advanced_video.py
```diff
@@ -0,0 +1,68 @@
...
+@author: drmccloy
+"""
+
+import numpy as np
+from expyfun import ExperimentController, fetch_data_file
+
+print(__doc__)
+
+
+movie_path = fetch_data_file('video/example-video.mp4')
```
Eric89GXL: op.join?

Eric89GXL: oh wait this is a URL, never mind

expyfun/visual/_visual.py
Eric89GXL: I'd say go with what works

Eric89GXL:
AppVeyor is unhappy about `wget`, you can look at whatever the code uses to get Miniconda installed to see how to download

drammock:
@Eric89GXL I am giving up on AVbin + appveyor.  The AVbin windows installer appears to not accept any command-line flags, and I've just spent about 2 hours trying to select the installer window and spoof an ENTER keypress through powershell, with no success.  Do we have a function decorator that will prevent a test from being run on windows?

Eric89GXL:
No such decorator. Probably we want to write one like `requires_avbin` that checks to see if the `avbin` executable is found and skips the test if not (on Linux, OSX, or Windows).

drammock:
@Eric89GXL you cool with this now?

expyfun/_utils.py
```diff
@@ -37,6 +37,14 @@
     pass
 
 try:
+    from pyglet.media import avbin
+    del avbin
+except ImportError:
+    has_avbin = False
+else:
+    has_avbin = True
```
Eric89GXL: So far we've managed to avoid importing `Pyglet` at all until a function that needs it is used. Could you make the decorator actually do this check internally? That way, this import is done only when running tests, not when running `expyfun`.

Eric89GXL: I know this bit is annoying, but I've seen imports fail for all sorts of reasons. Even doing `except Exception` above wouldn't necessarily take care of all cases, since Pyglet could do something catastrophic with memory. Even import problems aside, there is still import speed to consider. If importing Pyglet or its sub-libraries (including avbin) somehow becomes slow, now importing `expyfun` becomes slow, too.

Eric89GXL:
Also, when running `simple_video.py` I get:
```
  File "examples/stimuli/simple_video.py", line 24, in <module>
    ec.load_video(movie_path)
  File "/home/larsoner/Documents/python/expyfun/expyfun/_experiment_controller.py", line 758, in load_video
    self.video = Video(self, file_name, pos, units)
  File "/home/larsoner/Documents/python/expyfun/expyfun/visual/_visual.py", line 982, in __init__
    self._source = load(file_name)
  File "/home/larsoner/.local/lib/python2.7/site-packages/pyglet-1.2.2-py2.7.egg/pyglet/media/sources/loader.py", line 60, in load
  File "/home/larsoner/.local/lib/python2.7/site-packages/pyglet-1.2.2-py2.7.egg/pyglet/media/sources/loader.py", line 83, in load
  File "/home/larsoner/.local/lib/python2.7/site-packages/pyglet-1.2.2-py2.7.egg/pyglet/media/sources/riff.py", line 198, in __init__
pyglet.media.sources.riff.WAVEFormatException: AVbin is required to decode compressed media
```
This is because I don't have AVBin, but we should probably make it a more informative error (how should the user correct the problem?). Not necessarily for this PR though

Eric89GXL:
Other than my gripe about nesting the pyglet import in the decorator +1 for merge

expyfun/_utils.py
```diff
@@ -426,12 +418,21 @@ def verbose_dec(function, *args, **kwargs):
...
+        return False
+
+
 _is_appveyor = (os.getenv('APPVEYOR', 'False').lower() == 'true')
 requires_pandas = skipif(has_pandas is False, 'Requires pandas')
 requires_h5py = skipif(has_h5py is False, 'Requires h5py')
 requires_joblib = skipif(has_joblib is False, 'Requires joblib')
 requires_opengl21 = skipif(_is_appveyor, 'Appveyor OpenGL too old')
-requires_avbin = skipif(has_avbin is False, 'Requires AVbin')
+requires_avbin = skipif(_has_avbin() is False, 'Requires AVbin')
```
Eric89GXL: This will still be run on import. You will need to do something more like this, which only gets called when the decorator is accessed:

https://github.com/mne-tools/mne-python/blob/master/mne/utils.py#L680

drammock: this brings up another question: in the mne function you linked to, it says `skipif(not condition, ...)` whereas the functions in expyfun._utils say `skipif(condition is False, ...)`.  I was initially going to change those to `not condition` but decided you must have had a reason to do the more stringent `condition is False` test.  But seeing that mne-python does it the way I would have done it, I thought I'd raise the question.

Eric89GXL: Doesn't make a difference when dealing with bools here.

I started doing `not X` instead of `X is False` to allow for e.g. `0` vs `1` as parameters to behave more naturaly, because these evaluate in the expected way with `not X` (0->True, 1->False) whereas `X is False` will not (0->False, 1->False). The only time I do `X is True` or `X is False` now is when I actually want to differentiate e.g. `None`, `False`, `0`, `[]`, `''`, etc., which all pass the same way with `if not X` but will be different for `if X is False`.

drammock: OK, I won't bother changing, since in this case the genuine bools are the only possible type getting passed in, there's no need to allow for ints and strings to pass as truthy or falsey.

Eric89GXL:
Thanks @drammock

PR #276, opened 2016-03-22
[ENH] Keyboard key releases.

ilogue:
Collecting key-release events

~~I've added a `get_releases()` method on `ExperimentController` to be able to get key release events.~~

~~There's room for refactoring as I've just duplicated most of the keypress code.~~

~~I haven't been able to run/ adapt the tests properly since nosetest in a virtualenv uses global packages and I get the _check_pyglet error.~~ *resolved this with http://stackoverflow.com/a/864967* 
I have added an example script.

Summary of todo's:

- [x] fix docstring
- [x] fix event type argument
- [x] make press events dependent on argument
- [x] fix event tuple structure
- [x] fix demo
- [x] runtime error for TDT
- [x] EC tests

### DIFF
--- a/examples/experiments/keyrelease_demo.py
+++ b/examples/experiments/keyrelease_demo.py
@@ -12,8 +12,8 @@ inprecise timing.
 .. warning:
 
     It is currently not possible to get key-release events for Cedrus boxes or
-    TDT. Therefore, using get_presses(type='releases') or
-    get_presses(type='both') will throw an exception.
+    TDT. Therefore, using get_presses(kind='releases') or
+    get_presses(kind='both') will throw an exception.
 
 """
 # Author: Jasper van den Bosch <jasperb@uw.edu>
@@ -36,9 +36,9 @@ with ExperimentController('KeyPressAndReleaseDemo', screen_num=0,
     ec.wait_secs(isi)
 
     ###########################################
-    # listen_presses / while loop / get_presses(type='both')
+    # listen_presses / while loop / get_presses(kind='both')
     instruction = ("Press and release some keys\n\nlisten_presses()"
-                   "\nwhile loop {}\nget_presses(type='both')")
+                   "\nwhile loop {}\nget_presses(kind='both')")
     disp_time = wait_dur
     countdown = ec.current_time + disp_time
     ec.call_on_next_flip(ec.listen_presses)
@@ -51,7 +51,7 @@ with ExperimentController('KeyPressAndReleaseDemo', screen_num=0,
             # redraw text with updated disp_time
             ec.screen_text(instruction.format(disp_time))
             ec.flip()
-    events = ec.get_presses(type='both')
+    events = ec.get_presses(kind='both')
     ec.write_data_line('listen / while / get_presses', events)
     if not len(events):
         message = 'no keys pressed'
--- a/examples/experiments/keyrelease_demo.py
+++ b/examples/experiments/keyrelease_demo.py
@@ -1,10 +1,20 @@
 """
-=============
+=======================
 KeyPressAndRelease demo
-=============
+=======================
 
 This example demonstrates gathering key-releases as well as presses with
 the ExperimentController class.
+
+Please note that this currently only works for the keyboard, which has
+inprecise timing.
+
+.. warning:
+
+    It is currently not possible to get key-release events for Cedrus boxes or
+    TDT. Therefore, using get_presses(type='releases') or
+    get_presses(type='both') will throw an exception.
+
 """
 # Author: Jasper van den Bosch <jasperb@uw.edu>
 #
--- a/examples/experiments/keyrelease_demo.py
+++ b/examples/experiments/keyrelease_demo.py
@@ -1,0 +1,61 @@
+"""
+=============
+Keyrelease demo
+=============
+
+This example demonstrates the different keyrelease-gathering techniques available
+in the ExperimentController class.
+"""
+# Author: Dan McCloy <drmccloy@uw.edu>
+#
+# License: BSD (3-clause)
+
+import matplotlib.pyplot as plt
+
+from expyfun import ExperimentController
+import expyfun.analyze as ea
+
+print(__doc__)
+
+
+isi = 0.5
+wait_dur = 3.0
+msg_dur = 3.0
+
+with ExperimentController('KeyreleaseDemo', screen_num=0,
+                          window_size=[1280, 960], full_screen=False,
+                          stim_db=0, noise_db=0, output_dir=None,
+                          participant='foo', session='001',
+                          version='dev') as ec:
+    ec.wait_secs(isi)
+
+
+    ###########################################
+    # listen_releases / while loop / get_releases
+    disp_time = wait_dur
+    countdown = ec.current_time + disp_time
+    ec.call_on_next_flip(ec.listen_presses)
+    ec.screen_text('release some keys\n\nlisten_presses()'
+                   '\nwhile loop {}\nget_releases()'.format(disp_time))
+    ec.flip()
+    while ec.current_time < countdown:
+        cur_time = round(countdown - ec.current_time, 1)
+        if cur_time != disp_time:
+            disp_time = cur_time
+            # redraw text with updated disp_time
+            ec.screen_text('release some keys\n\nlisten_presses() '
+                           '\nwhile loop {}\nget_releases()'.format(disp_time))
+            ec.flip()
+    released = ec.get_releases()
+    ec.write_data_line('listen / while / get_releases', released)
+    if not len(released):
+        message = 'no keys released'
+    else:
+        message = ['{} released after {} secs\n'
+                   ''.format(key, round(time, 4)) for key, time in released]
+        message = ''.join(message)
+    ec.screen_prompt(message, msg_dur)
+    ec.wait_secs(isi)
+
+
+
--- a/examples/experiments/keyrelease_demo.py
+++ b/examples/experiments/keyrelease_demo.py
@@ -3,20 +3,16 @@
 Keyrelease demo
 =============
 
-This example demonstrates the different keyrelease-gathering techniques available
+This example demonstrates the keyrelease-gathering technique available
 in the ExperimentController class.
 """
 # Author: Dan McCloy <drmccloy@uw.edu>
 #
 # License: BSD (3-clause)
 
-import matplotlib.pyplot as plt
-
 from expyfun import ExperimentController
-import expyfun.analyze as ea
 
 print(__doc__)
-
 
 isi = 0.5
 wait_dur = 3.0
@@ -28,7 +24,6 @@ with ExperimentController('KeyreleaseDemo', screen_num=0,
                           participant='foo', session='001',
                           version='dev') as ec:
     ec.wait_secs(isi)
-
 
     ###########################################
     # listen_releases / while loop / get_releases
@@ -56,6 +51,3 @@ with ExperimentController('KeyreleaseDemo', screen_num=0,
         message = ''.join(message)
     ec.screen_prompt(message, msg_dur)
     ec.wait_secs(isi)
-
-
-
--- a/examples/experiments/keyrelease_demo.py
+++ b/examples/experiments/keyrelease_demo.py
@@ -31,7 +31,7 @@ with ExperimentController('KeyreleaseDemo', screen_num=0,
     countdown = ec.current_time + disp_time
     ec.call_on_next_flip(ec.listen_presses)
     ec.screen_text('release some keys\n\nlisten_presses()'
-                   '\nwhile loop {}\nget_releases()'.format(disp_time))
+                   '\nwhile loop {}\nget_presses()'.format(disp_time))
     ec.flip()
     while ec.current_time < countdown:
         cur_time = round(countdown - ec.current_time, 1)
@@ -39,15 +39,15 @@ with ExperimentController('KeyreleaseDemo', screen_num=0,
             disp_time = cur_time
             # redraw text with updated disp_time
             ec.screen_text('release some keys\n\nlisten_presses() '
-                           '\nwhile loop {}\nget_releases()'.format(disp_time))
+                           '\nwhile loop {}\nget_presses()'.format(disp_time))
             ec.flip()
-    released = ec.get_releases()
-    ec.write_data_line('listen / while / get_releases', released)
+    released = ec.get_presses(releases=True)
+    ec.write_data_line('listen / while / get_presses', released)
     if not len(released):
         message = 'no keys released'
     else:
-        message = ['{} released after {} secs\n'
-                   ''.format(key, round(time, 4)) for key, time in released]
+        message = ['{} {} after {} secs\n'
+                   ''.format(k, r, round(t, 4)) for k, t, r in released]
         message = ''.join(message)
     ec.screen_prompt(message, msg_dur)
     ec.wait_secs(isi)
--- a/examples/experiments/keyrelease_demo.py
+++ b/examples/experiments/keyrelease_demo.py
@@ -1,10 +1,10 @@
 """
 =============
-Keyrelease demo
+KeyPressAndRelease demo
 =============
 
-This example demonstrates the keyrelease-gathering technique available
-in the ExperimentController class.
+This example demonstrates gathering key-releases as well as presses with
+the ExperimentController class.
 """
 # Author: Jasper van den Bosch <jasperb@uw.edu>
 #
@@ -18,7 +18,7 @@ isi = 0.5
 wait_dur = 3.0
 msg_dur = 3.0
 
-with ExperimentController('KeyreleaseDemo', screen_num=0,
+with ExperimentController('KeyPressAndReleaseDemo', screen_num=0,
                           window_size=[1280, 960], full_screen=False,
                           stim_db=0, noise_db=0, output_dir=None,
                           participant='foo', session='001',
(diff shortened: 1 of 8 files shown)


examples/experiments/keyrelease_demo.py
```diff
@@ -0,0 +1,53 @@
+"""
+=============
+Keyrelease demo
+=============
+
+This example demonstrates the keyrelease-gathering technique available
+in the ExperimentController class.
+"""
+# Author: Dan McCloy <drmccloy@uw.edu>
```
drammock: should probably change this line :)

expyfun/_version.py
```diff
@@ -1 +1,2 @@
 __version__ = '2.0.0.dev0'
+__fork__ = 'ilogue'
```
drammock: don't change this file manually.  in the repo it only has `__version__` defined as `dev0`, but when you run `setup.py` or `expyfun.download_version()` it will get automatically altered to include a commit hash at the end of the version string, and in the case of `download_version()` will also add the __fork__ variable.

expyfun/_input_controllers.py
```diff
@@ -84,6 +99,17 @@ def _on_pyglet_keypress(self, symbol, modifiers, emulated=False):
             this_key = this_key.lstrip('_').lstrip('NUM_')
         self._keyboard_buffer.append((this_key, key_time))
 
+    def _on_pyglet_keyrelease(self, symbol, modifiers, emulated=False):
```
drammock: as far as I can tell, the definition of `_retrieve_keyboard_releases` and `_on_pyglet_keyrelease` are identical to the corresponding keypress methods.  That would mean that what you're actually returning is key press times, not key release times.  Am I missing something?

ilogue: @drammock  The only difference is that they use `self._keyboard_buffer_releases`, which contains only the release events.

expyfun/_version.py
ilogue: Yes this is the auto-generated line. Should I revert it?

expyfun/_input_controllers.py
drammock: But `self._keyboard_buffer_releases` is just a list you created... When you do `self.win.dispatch_events()` you're still going to get the same information from the window manager (which is press times, not release times) regardless of the name of the list you stick the values into.

Eric89GXL: I think that the `self.win.on_key_release = self._on_pyglet_keyrelease` should make it get the release events, no?

expyfun/_version.py
drammock: Yes. The recommended config for development is to install expyfun using `python setup.py develop` (instead of `install`) which avoids creating this line (and also should avoid the virtual environment testing issue, because the version seen by the system will be whatever branch you happen to have checked out)

Eric89GXL:
The real sticking point for me is that this will work for Pyglet only. It will not work for Cedrus boxes, or the TDT -- support would need to be added for those devices, too. For the Cedrus it might not be so bad (would have to look at what `pyxid` gives you) but for the TDT you'd have to modify the circuit.

Eric89GXL:
(And Pyglet keypresses are not designed to be high-accuracy, so having it only be there isn't of much benefit for experiments)

expyfun/_input_controllers.py
drammock: My mistake. This will work (for pyglet windows). We'll probably need to figure out how to get it working with TDT before merging it into master though. And possibly a refactor to make it more DRY.

expyfun/_input_controllers.py
```diff
@@ -489,3 +532,23 @@ def _retrieve_events(self, live_keys):
...
+    def _retrieve_keyboard_releases(self, live_keys):
+        # add escape keys
+        if live_keys is not None:
+            live_keys = [str(x) for x in live_keys]  # accept ints
+            live_keys.extend(self.force_quit_keys)
+        # pump for events
+        self._dev.poll_for_response()
+        while self._dev.response_queue_size() > 0:
+            key = self._dev.get_next_response()
+            if not key['pressed']:
```
ilogue: This is a way to implement it for `CedrusBox`: if `key['pressed']` is `False`, it's a release event. 

I don't know how to test this though. Do we have one of these at I-LABS?

drammock: we have one, but it's not working smoothly with expyfun yet.  I think @Eric89GXL was going to work on it during his next visit (?)  @rkmaddox knows more about it than I do.

rkmaddox: @Eric89GXL is planning to fix it when he's here next week.

Eric89GXL: I think that when you poll for events, *all* events get cleared from the `pyxid` buffer. So whenever polling for releases *or* presses, you'll need to append to our internal `self._keyboard_buffer_*` lists. This function will currently drop the presses, and the press function will drop the releases, right?

But yes, I'll be able to test this when I get to Seattle. The more annoying part will be modifying the TDT circuit. I'll probably have to do that, but it woudl be good if this were theoretically correct for Cedrus (and verified correct for Pyglet) by the time I'm there.

ilogue: Good point. I'll try to make it a singular event retrieval thing (although I'll have to figure out how target keys fit into that).

FYI, I just need this for keyboard right now, certainly not for TDT. This is for Alexis' preferential gazing experiment.

ilogue:
Okay so I was thinking how to refactor this with less repeated code; two options here I think
- the event tuples contain a third value indicating press vs. release
- we maintain two separate 'buffer' lists, but the methods update/read from both.

I imagine the methods would then take an argument with which the user can specify which events they want. Would the default be presses only?

Anyone want to weigh in on this? @drammock @Eric89GXL @rkmaddox

Eric89GXL:
> the event tuples contain a third value indicating press vs. release

The internal, private list for Cedrus (and Pyglet if necessary) can keep both types of events, sure.

> I imagine the methods would then take an argument with which the user can specify which events they want

I don't think we should change the existing user-facing functions at all. `get_presses` should stay, and the new calls you added `get_releases` can also stay. Separating the presses from releases can be done in private functions as necessary.

drammock:
Maybe internally store as a dict with keys `key`, `press`, and `release`?  I think I disagree with @Eric89GXL though: I would rather see a single `get_presses` function with kwargs for whether to return press time, release time, or both.

Eric89GXL:
>  I would rather see a single get_presses function

I can live with that

expyfun/_experiment_controller.py
```diff
@@ -933,6 +935,8 @@ def get_presses(self, live_keys=None, timestamp=True, relative_to=None):
             A time relative to which timestamping is done. Ignored if
             timestamp==False.  If ``None``, timestamps are relative to the time
             `listen_presses` was last called.
+        releases : False | bool
+            Whether key-release events should be returned as well.
```
drammock: I think the convention is to only include accepted *types* on the first line, not default values.  So it should read `releases : bool`.  Information about the default value goes in the following docstring line, so:

    Whether to include key-release events as well as key-press events (default ``False``).

expyfun/_experiment_controller.py
```diff
@@ -940,6 +944,8 @@ def get_presses(self, live_keys=None, timestamp=True, relative_to=None):
             If timestamp==False, returns a list of strings indicating which
             keys were pressed. Otherwise, returns a list of tuples
             (str, float) of keys and their timestamps.
+            If releases is True, the last value in the tuple will be a string,
+            either 'press' or 'release'.
```
drammock: @Eric89GXL what do you think about this?  I think putting `'press'` or `'release'` inside the return tuple is the right way to handle it, but I'm iffy on whether to have it only included when `releases=True`, versus always including it (so that all tuples would be `(<key>, <time>, 'press')` if `releases=False`).

ilogue: The reason I went for this initially, is that it won't break existing code that doesn't use releases.

If people have a 
```python
for key, time in presses:
    doSomething()
```
It would break if you added a third element. But if we don't mind then we could return a 3-tuple or a dict or even a `ResponseEvent` object.

drammock: we have a built-in `assert_version` function in expyfun that is tied to GitHub commit hashes, and a `download_version` method that puts the desired version of the codebase in the `./` directory relative to the script. So breaking old code is less of a danger if we assume that users are making use of these tools.  So I think we don't mind.  It's more a question about code complexity within the package; more standardized return signatures is better in my view because it means fewer conditionals elsewhere.

expyfun/_input_controllers.py
```diff
@@ -157,18 +167,29 @@ def check_force_quit(self, keys=None):
...
+                if releases:
+                    correctedEvent.append(event[2])
+                elif event[2] == 'release':
+                    continue  # removes releases.
+                if len(correctedEvent) > 1:
+                    correctedEvent = tuple(correctedEvent)
+                else:
+                    correctedEvent = correctedEvent[0]
+                correctedEvents.append(correctedEvent)
+        return correctedEvents
```
drammock: e.g., this passage I think could be simplified by standardizing the `get_presses` return signature.

expyfun/_experiment_controller.py
drammock: @Eric89GXL just convinced me that rather than `releases=False` (or `True`) we have a `type` parameter that takes a string value (`presses`, `releases`, or `both`).

expyfun/_experiment_controller.py
drammock: @Eric89GXL and I also agreed that it should always return a tuple (length 2 if `timestamp=False`, length 3 otherwise) and the last element of the tuple should be `'press'` or `'release'`.

expyfun/_experiment_controller.py
```diff
@@ -948,7 +954,7 @@ def get_presses(self, live_keys=None, timestamp=True, relative_to=None):
         ExperimentController.wait_for_presses
         """
         return self._response_handler.get_presses(live_keys, timestamp,
-                                                  relative_to)
+                                                  relative_to, releases)
```
drammock: all such cases will need to change to be `type` rather than `releases`

examples/experiments/keyrelease_demo.py
```diff
@@ -0,0 +1,53 @@
...
+with ExperimentController('KeyreleaseDemo', screen_num=0,
+                          window_size=[1280, 960], full_screen=False,
+                          stim_db=0, noise_db=0, output_dir=None,
+                          participant='foo', session='001',
+                          version='dev') as ec:
+    ec.wait_secs(isi)
+
+    ###########################################
+    # listen_releases / while loop / get_releases
+    disp_time = wait_dur
```
drammock: why the vacuous renaming here?

examples/experiments/keyrelease_demo.py
```diff
@@ -0,0 +1,53 @@
...
+    ec.screen_text('release some keys\n\nlisten_presses()'
+                   '\nwhile loop {}\nget_presses()'.format(disp_time))
+    ec.flip()
+    while ec.current_time < countdown:
+        cur_time = round(countdown - ec.current_time, 1)
+        if cur_time != disp_time:
+            disp_time = cur_time
+            # redraw text with updated disp_time
+            ec.screen_text('release some keys\n\nlisten_presses() '
+                           '\nwhile loop {}\nget_presses()'.format(disp_time))
```
drammock: this (combined with line 33) could be more DRY:

```
msg = 'release some keys\n\nlisten_presses()\nwhile loop {}\nget_presses()'
ec.screen_text(msg.format(disp_time))
```

examples/experiments/keyrelease_demo.py
```diff
@@ -0,0 +1,53 @@
...
+            ec.flip()
+    released = ec.get_presses(releases=True)
+    ec.write_data_line('listen / while / get_presses', released)
+    if not len(released):
+        message = 'no keys released'
```
drammock: this message will not happen if I press during the countdown but never release it.  So checking `len(released)` is testing whether any keys were *either pressed or released*.  This problem should go away, however, after implementing `get_presses(..., type='releases')` instead of the `releases=True` parameter.

expyfun/_experiment_controller.py
```diff
@@ -917,7 +918,8 @@ def listen_presses(self):
         """
         self._response_handler.listen_presses()
 
-    def get_presses(self, live_keys=None, timestamp=True, relative_to=None):
+    def get_presses(self, live_keys=None, timestamp=True, relative_to=None,
+                    releases=False):
```
drammock: The behavior that we want is that every time we call `get_presses` the entire pyglet / cedrus buffer gets cleared.  The docstring will need to make it very clear (possibly in a Notes section) that if you call `get_presses(..., type='releases')` then the press information will be discarded.  (similarly for `get_presses(..., type='presses')` then the releases will be discarded).  The reason for doing it this way is that the alternative is worse: if users only care about presses for a while, then at some point they start caring about releases, the first time they get releases there will be a big backlog of them from all the earlier unretrieved release events.

drammock:
For now, if `ExperimentController._response_device == 'tdt'`, we should throw a `RuntimeError` if users call `get_presses(..., type='releases')` or `type='both'`.

drammock:
So other than my copious comments, the only other thing missing is a few tests, which should probably go [around here](https://github.com/LABSN/expyfun/blob/master/expyfun/tests/test_experiment_controller.py#L197-L200)

examples/experiments/keyrelease_demo.py
ilogue: CTRL+H lazyiness :)

expyfun/_experiment_controller.py
ilogue: Also, I've been using [Google style docstrings](http://sphinxcontrib-napoleon.readthedocs.org/en/latest/example_google.html) as sphinx can autogenerate documentation from them. (But maybe that works with this format too?)

Eric89GXL: We follow the [numpy docstring standard] (https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) since that's what numpy, scipy, scikit-learn, mne-python, etc. follow

ilogue:
Note that `correct_presses` is currently also responsible for logging events. So if we filter out events before we pass then to `correct_presses`, they won't be logged. Is that intended?
However if we don't filter events in `_retrieve_events`, then a method like `_wait_for_one_press` will have to check for event type itself.

drammock:
I think it's acceptable to only log the types of events the user cares about. Especially since it's impossible to, e.g., press a key once and release it twice. Edge cases like: press and hold without releasing during the listening interval, can be handled via `type='both'`

ilogue:
Any remaining todo's? @drammock

expyfun/_experiment_controller.py
```diff
@@ -933,13 +936,17 @@ def get_presses(self, live_keys=None, timestamp=True, relative_to=None):
...
+            ``both``. (default ``presses``)
 
         Returns
         -------
         presses : list
-            If timestamp==False, returns a list of strings indicating which
-            keys were pressed. Otherwise, returns a list of tuples
-            (str, float) of keys and their timestamps.
+            Returns a list of tuples with key events. Each tuple's first value
+            will be the key pressed. If timestamp==True, the second value is
```
drammock: surround `timestamp=True` in double-backticks to get code-formatting in the rendered docs.  I go back and forth about whether to use the double-equals in this context; single equals also makes sense if we interpret it as part of the function signature, a la `keyword=value`.  But since I can't remember which way we've been doing it, I'm going to assume that we haven't always been consistent and won't enforce one or the other here.

expyfun/_input_controllers.py
```diff
@@ -61,19 +65,22 @@ def _clear_keyboard_events(self):
...
         if live_keys is not None:
             live_keys = [str(x) for x in live_keys]  # accept ints
             live_keys.extend(self.force_quit_keys)
         self.win.dispatch_events()  # pump events on pyglet windows
         targets = []
+
         for key in self._keyboard_buffer:
-            if live_keys is None or key[0] in live_keys:
-                targets.append(key)
+            if key[2] in self.keyEventTypes[type]:
```
drammock: I'm worried about the `[2]` index, so just to make sure: `self._keyboard_buffer` will always have 3-element tuples, right?  Because the filtering out of timestamps happens later?  If that's not the case, switching `[2]` to be `[-1]` might make sense.

drammock:
Other than that last nitpicky docstring comment, LGTM.  @Eric89GXL does this have your blessing?

expyfun/_input_controllers.py
ilogue: Yes, internally they will always be 3-tuples

ilogue:
Not sure why the AppVeyor build is failing;

```
Test EC mouse click support ... SKIP: Skipping test: test_mouse_clicks: Appveyor OpenGL too old
Test setting background color ... SKIP: Skipping test: test_background_color: Appveyor OpenGL too old
Test EL methods ... SKIP: Skipping test: test_eyelink_methods: Appveyor OpenGL too old
Test logging to file (Pyglet) ... Command exited with code -1073740771
```

drammock:
yeah, every once in a while it just fails on `test logging to file (pyglet)` and I don't know why either.  @Eric89GXL is the only one with the power to restart the appveyor jobs without doing an empty `push --force` to re-trigger the hook.

examples/experiments/keyrelease_demo.py
```diff
@@ -0,0 +1,53 @@
+"""
+=============
+KeyPressAndRelease demo
+=============
```
Eric89GXL: extend to same length as text

expyfun/_input_controllers.py
```diff
@@ -82,7 +89,12 @@ def _on_pyglet_keypress(self, symbol, modifiers, emulated=False):
             from pyglet.window import key
             this_key = key.symbol_string(symbol).lower()
             this_key = this_key.lstrip('_').lstrip('NUM_')
-        self._keyboard_buffer.append((this_key, key_time))
+        pressOrRelease = {True: 'press', False: 'release'}[isPress]
```
Eric89GXL: don't use camelCase for var names please

expyfun/_experiment_controller.py
```diff
@@ -948,7 +955,7 @@ def get_presses(self, live_keys=None, timestamp=True, relative_to=None):
         ExperimentController.wait_for_presses
         """
         return self._response_handler.get_presses(live_keys, timestamp,
-                                                  relative_to)
+                                                  relative_to, type)
```
Eric89GXL: Don't shadow builtins (don't use name `type`, perhaps `kind` instead)

Also, I don't see `type` value- or type-checked anywhere, so currently a user passing a typo `type='pressees` will instead get some ambiguous `KeyError` instead of a clear message about what values are accepted

expyfun/tests/test_experiment_controller.py
```diff
@@ -197,6 +197,15 @@ def test_ec(ac=None, rd=None):
...
         assert_equal(ec.get_presses(), [])
+        assert_equal(ec.get_presses(type='presses'), [])
+        if this_rd == 'tdt':
+            # TDT does not have key release events, so should raise an
+            # exception if asked for them:
+            assert_raises(RuntimeError, ec.get_presses, type='releases')
+            assert_raises(RuntimeError, ec.get_presses, type='both')
+        else:
+            assert_equal(ec.get_presses(type='both'), [])
+            assert_equal(ec.get_presses(type='releases'), [])
```
Eric89GXL: So hopefully eventually your changes should make it so that `assert_raises(ValueError, ec.get_presses, kind='foo')` will work

examples/experiments/keyrelease_demo.py
```diff
@@ -0,0 +1,53 @@
...
+
+This example demonstrates gathering key-releases as well as presses with
+the ExperimentController class.
```
Eric89GXL: Please add a `.. warning::` directive noting that it currently does not work on TDT or Cedrus boxes, only keyboard. Also please add a similar warning in the `get_presses` docstring. Since we don't guarantee accurate timing for anything other than TDT and Cedrus use, it's important that people know the timing isn't tested, and is probably bad, when using keyboard and/or Pyglet audio.

expyfun/_experiment_controller.py
ilogue: @drammock, are you OK with `kind` instead of `type`?

drammock: Sure.

PR #280, opened 2016-06-16
TDT delay

rkmaddox:
We have a TDT_DELAY parameter for delaying the audio and triggers so that the audio and video line up. If you want to present stimuli rapidly and you don't care about AV sync, this is a problem where that delay is large (as it is on the MEG computer). This allows you to set it.

NB: well after I spent a bunch of time on this, I realized this could be done by passing a dict as `audio_controller` when instantiating `ExperimentController`. I decided to finish it because it has two advantages: input type and value checking, and not requiring the TDT to be used, making development on your own computer easier.

### DIFF
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -12,6 +12,7 @@ import warnings
 from os import path as op
 from functools import partial
 import traceback as tb
+from numbers import Integral
 try:
     import pyglet
     from pyglet import gl
@@ -29,7 +30,6 @@ from ._input_controllers import Keyboard, CedrusBox, Mouse
 from .stimuli._filter import resample
 from .visual import Text, Rectangle, Video, _convert_color
 from ._git import assert_version
-
 
 # Note: ec._trial_progress has three values:
 # 1. 'stopped', which ec.identify_trial turns into...
@@ -93,8 +93,6 @@ class ExperimentController(object):
         default the mode is 'dummy', since setting up the parallel port
         can be a pain. Can also be a dict with entries 'type' ('parallel')
         and 'address' (e.g., '/dev/parport0').
-    verbose : bool, str, int, or None
-        If not None, override default verbose level (see expyfun.verbose).
     check_rms : str | None
         Method to use in checking stimulus RMS to ensure appropriate levels.
         Possible values are ``None``, ``wholefile``, and ``windowed`` (the
@@ -107,6 +105,14 @@ class ExperimentController(object):
         the expected version of the expyfun codebase is being used when running
         experiments. To override version checking (e.g., during development)
         use ``version='dev'``.
+    tdt_delay : int | None
+        The number of milliseconds to delay the audio signal so that it
+        coincides with the visual signal. Defualt is ``None``, which uses the
+        value from the configuration file (or sets it to 0 if that value is not
+        defined in the file). Otherwise it can be any non-negative integer.
+        This parameter has no effect if the TDT is not the audio device.
+    verbose : bool, str, int, or None
+        If not None, override default verbose level (see expyfun.verbose).
 
     Returns
     -------
@@ -126,13 +132,14 @@ class ExperimentController(object):
                  full_screen=True, force_quit=None, participant=None,
                  monitor=None, trigger_controller=None, session=None,
                  verbose=None, check_rms='windowed', suppress_resamp=False,
-                 version=None, enable_video=False):
+                 version=None, enable_video=False, tdt_delay=None):
         # initialize some values
         self._stim_fs = stim_fs
         self._stim_rms = stim_rms
         self._stim_db = stim_db
         self._noise_db = noise_db
         self._stim_scaler = None
+        self._tdt_delay = tdt_delay
         self._suppress_resamp = suppress_resamp
         self._enable_video = enable_video
         self.video = None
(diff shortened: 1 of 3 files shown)


drammock:
flake is mad, which makes Travis mad: `expyfun/tests/test_experiment_controller.py:418:80: E501 line too long (80 > 79 characters)`

Eric89GXL:
> flake is mad, which makes Travis mad

Which makes me mad. Get it right buddy.

rkmaddox:
Already fixed. Geez.

rkmaddox:
Also: input checking is THE WOOOORST. Why can't people just always give good inputs?

drammock:
flake [still mad](https://travis-ci.org/LABSN/expyfun/jobs/138170170#L690) but for different reasons.  

If you want to reeaaaally test your input checking, and make coveralls happier, [see here](https://coveralls.io/builds/6630845/source?filename=expyfun%2F_experiment_controller.py#L304)

Eric89GXL:
Maybe instead of input checking we can start writing "caveat salutor" in all our docstrings to save development time

expyfun/_experiment_controller.py
```diff
@@ -280,9 +287,34 @@ def __init__(self, exp_name, audio_controller=None, response_device=None,
             # Initialize devices
             #
 
+            # Handle the tdt_delay parameter
+            if self._tdt_delay is not None:
+                if self.audio_type == 'tdt':
+                    try:
+                        np.isfinite(self._tdt_delay)  # bombs if not numeric
+                    except:
```
Eric89GXL: never `except:` always `except Exception:` (the former even catches `SystemExit`). Also I think this might be good enough:
```
delay = int(tdt_delay)  # we can live with what this raises if it raises something
if delay < 0:
    raise ValueError(...)
```

Eric89GXL: (i.e., let `int()` do some of the heavy lifting on input checking)

Eric89GXL: ...and remove the `try` clause altogether

rkmaddox:
Did as @Eric89GXL suggested. If tests pass here, it's the last work I can put in before I move next week, and thus before Christina needs it.

rkmaddox:
pyflakes unhappy in 3.4. Looks like a problem with `_git.py`, which I did not change in this PR.

drammock:
that happened in another PR too.  If flake is the only error, I'd say go
ahead and consider this ready for review/merge.

rkmaddox:
Hold on, that might not be what the problem was. Let's see how these tests do.

Eric89GXL:
Don't fix [this flake error] (https://github.com/LABSN/expyfun/pull/281/commits/b75948d628be55d197df445b7824a1e089159705) if/when you hit it.

expyfun/_experiment_controller.py
```diff
@@ -107,6 +104,15 @@ class ExperimentController(object):
...
         use ``version='dev'``.
+    tdt_delay : int | None
+        The number of milliseconds to delay the audio signal so that it
+        coincides with the visual signal. Defualt is ``None``, which uses the
+        value from the configuration file (or sets it to 0 if that value is not
+        defined in the file). Otherwise it can be any non-negative integer.
+        Floats will be rounded. This parameter has no effect if the TDT is not
+        the audio device.
+    verbose : bool, str, int, or None
+        If not None, override default verbose level (see expyfun.verbose).
```
Eric89GXL: `verbose` wasn't actually moved in the arg list

Eric89GXL:
> I realized this could be done by passing a dict as audio_controller when instantiating ExperimentController

We should not have two ways of doing the same thing if it can be avoided.

> I decided to finish it because it has two advantages: input type and value checking

This can be moved to `_tdt_controller.py` to do the same thing.

> and not requiring the TDT to be used, making development on your own computer easier.

This change can also be made in `_tdt_controller.py` and/or `_experiment_controller.py`, not sure which. Basically any items in the `audio_controller=dict(...)` dict should override defaults that are pulled from the config. So if you do `audio_controller=dict(TDT_DELAY=0)` it should be made to work on both your local machine (where it will be ignored) and on the MEG computer where it will be used properly (along with the config-file defaults for TDT_MODEL, etc.).

Make sense?

expyfun/tests/test_experiment_controller.py
```diff
@@ -399,3 +399,32 @@ def test_background_color():
...
+    with ExperimentController(*std_args, tdt_delay=1, audio_controller='tdt',
+                              **std_kwargs) as ec:
+        ec.wait_secs(0)
+    with ExperimentController(*std_args, tdt_delay=None,
+                              audio_controller='tdt', **std_kwargs) as ec:
+        ec.wait_secs(0)
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter('always')
+        with ExperimentController(*std_args, tdt_delay=3,
+                                  audio_controller='pyglet',
```
drammock: trailing whitespace here is making travis flake out.  (my bad, but probably easier if @rkmaddox fixes)

rkmaddox:
@Eric89GXL it makes sense, but it's not quite clear to em how it would work. If you pass that dict, it sets the audio controller to be TDT. But if you don't have the TDT, it will use dummy mode. It seems like the dict should be a separate argument, rather than being an option for `audio_controller`, such that you still have the option of specifying TDT, sound card, or system default.

Eric89GXL:
>  If you pass that dict, it sets the audio controller to be TDT

No, it only does that if you do `audio_controller=dict(TYPE='tdt', ...)`

rkmaddox:
Ohhhhhhhh. Well that was a waste of time.

--
Ross Maddox, Ph.D.
Postdoctoral Fellow
University of Washington
phone: 206-685-4662

Eric89GXL:
@drammock do you want to take over and move the type checking to `_tdt_controller.py`? I can do it if not

drammock:
I won't likely get to it in the next few days, so if you want to tackle it
while the problem is fresh, be my guest @Eric89GXL

Eric89GXL:
Closing for #283

PR #287, opened 2016-07-26
fix: windows GL error message handling

### LINKED ISSUE #286: set_background_color error on windows machine
Running any function from visuals (set_background_color, FixationDot, etc) on windows results in the following error:

```
  File "C:\Experiments\mbagheri\aud_att_int_SAMPLE\aud_att_int.py", line 35, in <module>
    ec.set_background_color([0.2] * 3)
  File "C:\Experiments\mbagheri\aud_att_int_SAMPLE\expyfun\_experiment_controller.py", line 540, in set_background_color
    Rectangle(self, pos=[0, 0, 2.1, 2.1], fill_color=color).draw()
  File "C:\Experiments\mbagheri\aud_att_int_SAMPLE\expyfun\visual\_visual.py", line 476, in __init__
    line_loop=True)
  File "C:\Experiments\mbagheri\aud_att_int_SAMPLE\expyfun\visual\_visual.py", line 191, in __init__
    _check_log(vertex, gl.glGetShaderInfoLog)
  File "C:\Experiments\mbagheri\aud_att_int_SAMPLE\expyfun\visual\_visual.py", line 172, in _check_log
    raise RuntimeError(message)
RuntimeError: No errors.
``` 
Updating pyglet didn't help. Any ideas?

drammock:
closes #286

### DIFF
--- a/expyfun/visual/_visual.py
+++ b/expyfun/visual/_visual.py
@@ -168,7 +168,9 @@ def _check_log(obj, func):
     ptr = cast(pointer(log), POINTER(c_char))
     func(obj, 4096, pointer(c_int()), ptr)
     message = log.value
-    if message:
+    if message.startswith('No errors'):
+        pass
+    elif message:
         raise RuntimeError(message)
 
 
--- a/expyfun/visual/_visual.py
+++ b/expyfun/visual/_visual.py
@@ -168,7 +168,7 @@ def _check_log(obj, func):
     ptr = cast(pointer(log), POINTER(c_char))
     func(obj, 4096, pointer(c_int()), ptr)
     message = log.value
-    if message.startswith('No errors'):
+    if message.startswith(b'No errors'):
         pass
     elif message:
         raise RuntimeError(message)


drammock:
Travis is mad because of unrelated Flake errors (when did those slip in?) and because of [a py3k string/bytes problem](https://travis-ci.org/LABSN/expyfun/jobs/147350277#L827).  Will try to iron these out tomorrow morning, if someone else doesn't get to it first.

Eric89GXL:
Restart the builds now it should work again:

https://gitlab.com/pycqa/flake8/issues/177#note_13316550

Eric89GXL:
...although I did restart and still saw errors, so I commented again:

https://gitlab.com/pycqa/flake8/issues/177#note_13318620

drammock:
@Eric89GXL travis fails are now all flake-related, so I think this is ready to merge.

expyfun/visual/_visual.py
```diff
@@ -168,7 +168,9 @@ def _check_log(obj, func):
     ptr = cast(pointer(log), POINTER(c_char))
     func(obj, 4096, pointer(c_int()), ptr)
     message = log.value
-    if message:
+    if message.startswith(b'No errors'):
+        pass
+    elif message:
```
Eric89GXL: why not just `if not message.startswith(b'No errors')`?

Eric89GXL: ...and actually, to be backward compatible you almost certainly need to do `if message and not message.startswith(b'No errors'):`

Eric89GXL: (compared to what I had anyway)

Eric89GXL:
This should work and it's a show-stopper so I'll merge, but we should unify those checks at some point

Eric89GXL:
Thanks @drammock

drammock:
I think there is some cross-platform inconsistency here regarding what `message` is.  Testing on linux seems to yield `''` which won't choke on the `.startswith`, so as long as `message` is not `None` we should be fine.  I agree though that your second suggestion is probably better.  @mirbagheri said he came across some info online claiming that `'No errors'` actually means "there were some errors that we've buried in some other log file"...  I've asked him to add a link to that here.

mirbagheri:
According [here](http://webcache.googleusercontent.com/a/30432) the output of the compiler in `glGetShaderInfoLog` or 'glGetProgrammInfoLog' is driver-specific, and `'No error'` is only for Nvidia GPUs.
[Here](https://github.com/google/gxui/issues/1) people talk about this `COMPILE_STATUS` test as a workaround for the inconsistency.

drammock:
@mirbagheri first link is dead.

mirbagheri:
Link edited^

Eric89GXL:
So @mirbagheri we should be safe with the now-implemented check?

PR #288, opened 2016-07-26
WIP: Fix flake

Eric89GXL:
This should make the CIs happy, but we should wait to see if `flake8` folks fix it at their end:

https://gitlab.com/pycqa/flake8/issues/177#note_13318620

### DIFF
--- a/expyfun/_utils.py
+++ b/expyfun/_utils.py
@@ -37,21 +37,21 @@ except Exception:
     pass
 
 try:
-    import pandas  # noqa, analysis:ignore
+    import pandas  # noqa
 except ImportError:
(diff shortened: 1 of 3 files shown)


drammock:
I think this is no longer needed in light of https://gitlab.com/pycqa/flake8/commit/f82b5d62d0c4e48b95466bb259f3401aecf28de7

OK to close, or do we want to leave open until Travis starts using the more up-to-date flake?

Eric89GXL:
There were one or two legitimate ones in there I think, I need to revise it

Eric89GXL:
Nope, they were all fine

PR #294, opened 2017-01-10
MRG: Add sample rate test script

Eric89GXL:
Sample rate testing script. @rkmaddox can you see if you're happy?


examples/sync/plot_sample_rate_test.py
```diff
@@ -0,0 +1,60 @@
...
+======================
+Audio sample rate test
+======================
+
+This example tests the TDT sample rate, which we expect to be ``24414.0625``.
+To test this:
+
+1. Connect the TDT audio output to the sound card input. This can be on
+   the same machine that the TDT is connected to, or a different one.
+   A 1/8" male-male cable running from the headphone monitor to the
```
rkmaddox: to the...recording computer. Note that you will also need a 1/4" male to 1/8" female adapter to plug it into the RP2 / RP2.1 / RZ6 (whereas the RM1 has an 1/8" jack).

examples/sync/plot_sample_rate_test.py
```diff
@@ -0,0 +1,60 @@
...
+2. Start Audacity on the sound-card machine.
+3. Configure the sound-card machine and/or Audacity to record from the sound
+   card input.
+4. Tell Audacity to record.
+5. Run this script. It should take about 40 seconds.
```
rkmaddox: 40 seconds if the sampling rate is ~24414 (some folks run at 48828).

examples/sync/plot_sample_rate_test.py
```diff
@@ -0,0 +1,60 @@
...
+8. The sound that was played put 1e6 (1,000,000) samples between the two
+   peaks. So you can get the effective sample rate as:
+
+   .. math::
+
+       f_s = \frac{1000000}{t_{stop} - t_{start}}
+
+   For example, Eric's RM1 (2017/01/10) had a start time of ``19.854330`` sec
+   and an end time of ``60.813690`` sec for a difference of of ``40.95936``,
+   yielding an effective sample rate of ``24414.44`` Hz.
```
rkmaddox: This is terrible. We need to do this with every TDT we use to record ABRs. Even the difference between 24414.0625 (the stated sampling rate) and 24414.44 is enough to get off by 1 ms by the end of a 60 ms recording, which is enough to lose components of the ABR.

examples/sync/plot_sample_rate_test.py
```diff
@@ -0,0 +1,60 @@
...
+# License: BSD (3-clause)
+
+import numpy as np
+
+from expyfun import ExperimentController, building_doc
+
+print(__doc__)
+
+stim = np.zeros(int(1e6) + 1)
+stim[[0, -1]] = 1.
```
rkmaddox: I customarily add a zero to the end of all my stimuli since the TDT will stick at the final value if it's stopped there. Doesn't matter here since stimulus is not looped.

Eric89GXL: We fixed this in software already (it auto-appends a zero effectively.)

examples/sync/plot_sample_rate_test.py
```diff
@@ -0,0 +1,60 @@
...
+stim_dur = len(stim) / 24414.
+with ExperimentController('FsTest', full_screen=False, noise_db=-np.inf,
+                          participant='s', session='0', output_dir=None,
+                          suppress_resamp=True, check_rms=None,
+                          version='dev') as ec:
+    ec.identify_trial(ec_id='', ttl_id=[0])
+    ec.load_buffer(stim)
+    print('Starting stimulus')
+    ec.start_stimulus()
+    print('Stimulus started')
```
rkmaddox: ``print('Stimulus started. Please wait %i seconds.' % (stim_dur + 1))`` ?

rkmaddox:
Other than a couple comments LGTM. Thanks for adding.

examples/sync/plot_sample_rate_test.py
```diff
@@ -0,0 +1,60 @@
...
+
+import numpy as np
+
+from expyfun import ExperimentController, building_doc
+
+print(__doc__)
+
+stim = np.zeros(int(1e6) + 1)
+stim[[0, -1]] = 1.
+stim_dur = len(stim) / 24414.
```
drammock: Shouldn't this `stim_dur` definition be inside the `with ExperimentController(...) as ec` context, and use `ec.fs` as the divisor?

rkmaddox: Yeah, that's better.

examples/sync/plot_sample_rate_test.py
rkmaddox: Great.

Eric89GXL:
Okay changes pushed. Feel free to merge if you're happy @rkmaddox or @drammock

examples/sync/plot_sample_rate_test.py
```diff
@@ -44,17 +45,17 @@
...
                           suppress_resamp=True, check_rms=None,
                           version='dev') as ec:
     ec.identify_trial(ec_id='', ttl_id=[0])
     ec.load_buffer(stim)
-    print('Starting stimulus')
+    print('Starting stimulus.')
     ec.start_stimulus()
-    print('Stimulus started')
+    wait_dur = len(stim) / ec.fs + 1.
+    print('Stimulus started. Please wait %d seconds' % wait_dur)
```
rkmaddox: Add a period after seconds? May be the least consequential comment I've ever made.

Eric89GXL:
Worth it. Pushed.

rkmaddox:
+1 for merge when tests pass.

Eric89GXL:
Thanks for the reviews. Now time to test.

I can't currently test the RP2 I have here because I am missing cables but those are on order. Once they're here I'll have numbers for it.

We could put these numbers in the repo wiki, WDYT?

rkmaddox:
Yeah, I am dying to know about the RP2.1s in Seattle. I think they must be pretty damn close to 24414.0625, because my ABRs there look good, but it now seems possible that they might improve with real measurements, since a difference of 1 sample / s could effect the responses.

I have a couple cables on order from BrainVision that will allow me test RP2 --> actiChamp which is the true test needed to get the right rate for analysis.

Eric89GXL:
Cool, let us know how it goes. And if you end up needing to change or make new test scripts let's add them to `examples/sync/`

PR #295, opened 2017-01-24
MRG: Add RT calculation

Eric89GXL:
I've been meaning to add RT calculation to `press_times_to_hmfc` for a long time and finally got around to it. WIP because I want to see if I can get the OSX CI working.

### DIFF
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -12,11 +12,6 @@ import warnings
 from os import path as op
 from functools import partial
 import traceback as tb
-try:
-    import pyglet
-    from pyglet import gl
-except Exception:  # maybe we don't have a display (e.g., in terminal)
-    pyglet = gl = None
 
 from ._utils import (get_config, verbose_dec, _check_pyglet_version, wait_secs,
                      running_rms, _sanitize, logger, ZeroClock, date_str,
@@ -219,6 +214,7 @@ class ExperimentController(object):
             if screen_num is None:
                 screen_num = int(get_config('SCREEN_NUM', '0'))
             if monitor is None:
+                import pyglet
                 mon_size = pyglet.window.get_platform().get_default_display()
                 mon_size = mon_size.get_screens()[screen_num]
                 mon_size = [mon_size.width, mon_size.height]
@@ -533,6 +529,7 @@ class ExperimentController(object):
         ``glClearColor`` will be set so the buffer starts out with the
         appropriate backgound color.
         """
+        from pyglet import gl
         if not self._enable_video:
             # we go a little over here to be safe from round-off errors
             Rectangle(self, pos=[0, 0, 2.1, 2.1], fill_color=color).draw()
@@ -716,6 +713,7 @@ class ExperimentController(object):
             N x M x 3 array of screen pixel colors.
         """
         # next line must be done in order to instantiate image_buffer_manager
+        import pyglet
         data = pyglet.image.get_buffer_manager().get_color_buffer()
         data = self._win.context.image_buffer_manager.color_buffer.image_data
         data = data.get_data(data.format, data.pitch)
@@ -770,6 +768,8 @@ class ExperimentController(object):
 
 # ############################### OPENGL METHODS ##############################
     def _setup_window(self, window_size, exp_name, full_screen, screen_num):
+        import pyglet
+        from pyglet import gl
         # Use 16x sampling here
         config_kwargs = dict(depth_size=8, double_buffer=True, stereo=False,
                              stencil_size=0, samples=0, sample_buffers=0)
@@ -852,6 +852,7 @@ class ExperimentController(object):
         `call_on_next_flip`, followed by functions added with
         `call_on_every_flip`.
         """
+        from pyglet import gl
         if when is not None:
             self.wait_until(when)
         call_list = self._on_next_flip + self._on_every_flip
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -712,8 +712,8 @@ class ExperimentController(object):
         scr : array
             N x M x 3 array of screen pixel colors.
         """
-        # next line must be done in order to instantiate image_buffer_manager
         import pyglet
+        # this must be done in order to instantiate image_buffer_manager
         data = pyglet.image.get_buffer_manager().get_color_buffer()
         data = self._win.context.image_buffer_manager.color_buffer.image_data
         data = data.get_data(data.format, data.pitch)
--- a/expyfun/_sound_controllers.py
+++ b/expyfun/_sound_controllers.py
@@ -5,10 +5,12 @@
 #
 # License: BSD (3-clause)
 
-import numpy as np
-from scipy import fftpack
 import sys
 import os
+import warnings
+
+import numpy as np
+
 import pyglet
 
 _use_silent = (os.getenv('_EXPYFUN_SILENT', '') == 'true')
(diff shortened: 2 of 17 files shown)


Eric89GXL:
Okay this one should be good to go.

@drammock do you have time to look?

expyfun/analyze/_analyze.py
```diff
@@ -42,39 +54,67 @@ def press_times_to_hmfc(presses, targets, foils, tmin, tmax,
...
-        assert locs.min() >= 1
-
-    # figure out which presses were to target or masker (valid_idx)
-    in_window = (presses <= all_times[locs - 1] + dur)
-    valid_idx = np.where(in_window)[0]
-    n_other = np.sum(~in_window)
-
+    # figure out what targ/mask times our presses correspond to
+    press_to_stim = np.searchsorted(stim_times, presses, 'right') - 1
+    if len(press_to_stim) > 0:
```
drammock: don't need the `> 0` part

expyfun/analyze/_analyze.py
```diff
@@ -216,17 +256,18 @@ def wrapper(*args):
...
     Parameters
     ----------
     x : array-like
         Reaction time data to fit.
-
     axis : int | None
         The axis along which to calculate the chi-square fit. If none, ``x``
         will be flattened before fitting.
+    warn : bool
+        If True, warn about possible bad reaction times.
```
drammock: thanks for this.  The criterion for warning is too stringent, and always blows up my console.

expyfun/analyze/_analyze.py
```diff
@@ -254,7 +295,8 @@ def rt_chisq(x, axis=None):
     if axis is None:
         df, _, scale = ss.chi2.fit(x, floc=0)
     else:
-        fit = partial(ss.chi2.fit, floc=0)
+        def fit(x):
+            return np.array(ss.chi2.fit(x, floc=0))
```
drammock: are we not doing `partial` anymore?

expyfun/_experiment_controller.py
```diff
@@ -12,11 +12,6 @@
 from os import path as op
 from functools import partial
 import traceback as tb
-try:
-    import pyglet
-    from pyglet import gl
-except Exception:  # maybe we don't have a display (e.g., in terminal)
-    pyglet = gl = None
```
Eric89GXL: It should make `import expyfun` work even on systems without a display (i.e., the OSX Travis CI). And it has the advantage of making all `expyfun` sub-imports, like to `expyfun.analyze` or `stimuli` faster. It only slows down and creates an OpenGL context when necessary, i.e. when instantiating `ExperimentController`

expyfun/_sound_controllers.py
```diff
@@ -5,10 +5,12 @@
...
 # License: BSD (3-clause)
 
-import numpy as np
-from scipy import fftpack
 import sys
 import os
+import warnings
+
+import numpy as np
+
```
Eric89GXL: It's a convention borrowed from MNE to separate standard imports, then standard scientific library imports, then local imports

expyfun/analyze/_analyze.py
Eric89GXL: It won't work with newest `numpy` that requires that whatever is passed to `apply_along_axis` give a `np.ndarray` out instead of, in this case, a tuple

expyfun/_experiment_controller.py
drammock: rad.

expyfun/analyze/_analyze.py
drammock: dang. OK.

Eric89GXL:
Fixed the comment @drammock

expyfun/analyze/_analyze.py
```diff
@@ -42,39 +54,67 @@ def press_times_to_hmfc(presses, targets, foils, tmin, tmax,
...
-        assert locs.max() < len(all_times)  # should be True b/c of np.inf
-        assert locs.min() >= 1
-
-    # figure out which presses were to target or masker (valid_idx)
-    in_window = (presses <= all_times[locs - 1] + dur)
-    valid_idx = np.where(in_window)[0]
-    n_other = np.sum(~in_window)
-
+    # figure out what targ/mask times our presses correspond to
+    press_to_stim = np.searchsorted(stim_times, presses, 'right') - 1
```
drammock: is `right` vs. `left` going to make a difference here?  Should only matter if there are multiple elements of `stim_times` that are the same as each other and also happen to exactly match an element of `presses`.

Eric89GXL: We had it before so I kept it, but I guess not

Eric89GXL: removed it in latest commit

expyfun/analyze/_analyze.py
```diff
@@ -42,39 +54,67 @@ def press_times_to_hmfc(presses, targets, foils, tmin, tmax,
...
+    stim_times = stim_times[valid_mask]
+    order = order[valid_mask]
+    del valid_mask
+    press_to_stim, used_map_idx = np.unique(press_to_stim, return_index=True)
+    presses = presses[used_map_idx]
+    stim_times = stim_times[used_map_idx]
+    order = order[used_map_idx]
+    assert len(presses) == len(stim_times)
+    diffs = presses - (stim_times - tmin)
+    assert all(diffs >= tmin)
```
drammock: do we really want to be subtracting `tmin` here when calculating `diffs`?  If `stim_times` is 1, `tmin` is 0.2, and `presses` is 1.1, then the press should not count, but `diffs` would be 0.3 and `assert diffs >= tmin` would pass.

Eric89GXL: see above, `stim_times` has been adjusted by `tmin` so it needs to be adjusted back here

Eric89GXL: or to be more precise, `targets` and `foils` get adjusted by `tmin` and then concatenated into `stim_times`, so we need to adjust back here

expyfun/analyze/_analyze.py
```diff
@@ -42,39 +55,67 @@ def press_times_to_hmfc(presses, targets, foils, tmin, tmax,
...
+    assert all(diffs <= tmax)
+    del used_map_idx
     # figure out which of valid presses were to target or masker
-    used = np.unique(locs[valid_idx])  # unique to remove double-presses
-    orig_places = (inv_order[used - 1] - 1)
-    n_hit = sum(orig_places < len(targets))
-    n_fa = len(used) - n_hit
+    target_mask = (order <= len(targets))
```
drammock: I'm assuming that `-np.inf` is gone from `stim_times` by this point, so `0` is likewise gone from `order`?  If not, then even if there are zero hits, there will still be one `True` value in `target_mask`.  (By the way, the variable name `target_mask` is a bit confusing because of "target/masker").

Eric89GXL: Maybe the most constructive thing to do is to come up with examples that you think will break the code and test them. It will be more productive than chatting about I think. You can make a PR into my branch with more test cases, passing or failing, if you want?

drammock: fair point.  I tested this one and it works.

expyfun/analyze/_analyze.py
```diff
@@ -42,39 +55,67 @@ def press_times_to_hmfc(presses, targets, foils, tmin, tmax,
...
+    press_to_stim = np.searchsorted(stim_times, presses) - 1
+    if len(press_to_stim) > 0:
+        assert press_to_stim.max() < len(stim_times)  # True b/c of np.inf
+        assert press_to_stim.min() >= 0
+    stim_times = stim_times[press_to_stim]
+    order = order[press_to_stim]
+    assert (stim_times <= presses).all()
+
+    # figure out which presses were to target or masker (valid_mask)
+    valid_mask = (presses <= stim_times + dur)
```
drammock: this line is wrong, I think.  Should be `np.logical_and((stim_times + tmin <= presses), (presses <= stim_times + tmax))`.  Currently it's testing for presses being `(tmax-tmin)` seconds after stim time.

Eric89GXL:
See my previous comments about how adjustment by tmin that has already
occurred earlier.

Can you try to break it with an example?

I'll fix the unrelated float arithmetic case of your PR soon.

drammock:
arrgh, sorry, I missed those earlier responses about adding `tmin` into `targets` and `foils`...  I had been looking through the source to see if you'd done something like that, but didn't see it.  +1 for merge after the float issue is resolved.

Eric89GXL:
Corner case fixed (alongside suitable docstring warning) in https://github.com/LABSN/expyfun/pull/295/commits/6e3802769ce116f515a5c351fc065b0e9d83bf6f

drammock:
thanks @Eric89GXL .  I saw you changed the logic of the RT calcs to make the code more drammock-friendly, thanks for that. :)

Eric89GXL:
> I saw you changed the logic of the RT calcs

Yeah it made it a *little* bit less efficient (extra additions / subtractions) but the gain in readability is certainly worth it

PR #296, opened 2017-01-24
MRG: Add RP2 legacy mode

Eric89GXL:
@rkmaddox can you see if `'RP2legacy'` as your TDT_TYPE works with this circuit?

All I did was change the memory buffer sizes from 4e006 to 2e006, which I think is what we talked about working.


drammock:
cc @baumers

Eric89GXL:
@rkmaddox can you verify this works with your RP2? If so then we should be good for merge.

rkmaddox:
Huh, thought I tested already. Will do this afternoon.

Eric89GXL:
Actually you might have already, and just responded via email. Let me check.

Eric89GXL:
Yep, on Feb 2nd you made sure it worked. @drammock feel free to merge if you're happy, then.

rkmaddox:
Well that was easy.

drammock:
@baumers we should talk about this (offline) in regards to testing the other RP2's ...  I forget where we left off with that.

baumers:
Sounds good! I have some notes but haven't bit focused on old RP2's for a couple weeks, I think I'm behind on updating a couple things on the wiki that I can take care of tomorrow.

PR #299, opened 2017-01-26
move stimgen script back into stimuli folder; add empty init

### LINKED ISSUE #297: simple_experiment broken
tries to do `from stimuli.plot_generate_simple_stimuli import generate_stimuli` and fails, because `plot_generate_simple_stimuli.py` is not in the `stimuli` subfolder (it's in the root of `examples`).

drammock:
closes #297.


Eric89GXL:
CircleCI is broken can you investigate?

drammock:
Here's the error:
```
  File "/home/ubuntu/expyfun/expyfun/stimuli/__init__.py", line 5, in <module>
    from ._hrtf import convolve_hrtf
  File "/home/ubuntu/expyfun/expyfun/stimuli/_hrtf.py", line 6, in <module>
    from ..io import read_hdf5
ValueError: Attempted relative import beyond toplevel package
```
It happened during execution of `examples/plot_simple_experiment.py`, which wasn't changed in this PR and doesn't call `hrtf` anywhere.  I think it happens because `plot_simple_experiment.py` does `from stimuli.plot_generate_simple_stimuli import generate_stimuli`, which triggers `from expyfun.stimuli import play_sound`, which triggers `expyfun.stimuli.__init__` which triggers `_hrtf`.

drammock:
this may have gone undetected for a while because the relative import that this PR changes doesn't get triggered if `equally_spaced_sinewaves.hdf5` already exists.  Is that possible?  In any case I'm not certain how to debug this locally, so it will have to wait a day.  Apparently CircleCI has a "rerun with SSH" option where you get 30 minutes to SSH in and view the full traceback logs...  first time dealing with CircleCI / sphinx / etc (that was always your baby)

Eric89GXL:
That's fine, no rush from my end. If you get sick of trying to fix it I can
take over

drammock:
@larsoner I've done a bit of reading.  The problem is that relative imports don't work properly when running the examples.  The solution I've seen in various places is to use the `-m` flag like `python -m app.sub1.mod1` when running a module file directly (as opposed to importing it).  Strangely, that's not what we're doing (we're running a non-module file in the `examples` directory and importing various expyfun modules/functions from within it).  So I don't really see how the recommended solution will fix this, nor do I understand why we're getting the error in the first place.  However, I can confirm (see my last commit) that doing an absolute import like `from expyfun.io import foo` rather than `from ..io import foo` fixes the problem.  Having fixed that one, sphinx immediately choked on the same type of relative import (this time from `..utils`, within `_mls.py`), which gets triggered by the next line down of `expyfun/stimuli/__init__.py`.  Traceback from CircleCI:

```
________________________________________________________________________________
/home/ubuntu/expyfun/examples/plot_simple_experiment.py failed to execute correctly:Traceback (most recent call last):
  File "/home/ubuntu/expyfun/sphinx-gallery/sphinx_gallery/gen_rst.py", line 475, in execute_code_block
    exec(code_block, example_globals)
  File "<string>", line 31, in <module>
  File "/home/ubuntu/expyfun/expyfun/stimuli/__init__.py", line 6, in <module>
    from ._mls import compute_mls_impulse_response, repeated_mls
  File "/home/ubuntu/expyfun/expyfun/stimuli/_mls.py", line 11, in <module>
    from .._utils import verbose_dec, logger
ValueError: Attempted relative import beyond toplevel package
________________________________________________________________________________
```

I'm wondering if maybe I've just done something dumb here?  Or is Sphinx being dumb?  It's as if the `__init__.py` file in stimuli was never getting run previously during testing, and now it is, and a bunch of errors are coming to light.  Thoughts / help?

Eric89GXL:
> The problem is that relative imports don't work properly when running the examples.

I think there might be some confusion about what imports we're talking about. There are expyfun *module* imports (like the one you changed in `expyfun/stimuli/_hrtf.py`) -- these have always worked as relative imports, and should stay that way for various reasons. They should not be the root of your problem.

Then there are imports within *example scripts* that try to load functions from other *example scripts* (some of which we might treat as their own little modules). We probably need to fix these.

>  The solution I've seen is to use the -m flag like python -m app.sub1.mod1 when running a module file directly (as opposed to importing it).

We shouldn't be running any module-level file. Just the `examples/*` files, in some appropriate manner.

> doing an absolute import like from expyfun.io import foo rather than from ..io import foo fixes the problem

We don't want to, and shouldn't need to restructure anything under `expyfun/`.

I wonder if some light could be shed on these problems by renaming the `examples/stimuli` folder to e.g. `examples/stimulus_generation` to eliminate the confusion with `expyfun/stimuli`, at least for testing purposes.

I suspect that the easiest solution might involve using `importlib` or `__import__` from within the script to import something from a particular directory (similar to what could be achieved by modifying `sys.path` and calling `import` directly).

Eric89GXL:
@drammock should I take this over, or do you still want to work on it?

drammock:
I would like it to work, and I would like to *understand* why it's not working, but I don't have any special attachment to being the one who fixes it.  Perhaps reviewing your attempts at this PR will be enough for me to understand it.  So: have at it :smile:

Eric89GXL:
Will take over

PR #301, opened 2017-03-10
MRG: Adaptive Tracker

rkmaddox:
Adds three classes to `expyfun.stimuli`: `TrackerUD`, `TrackerBinom`, and `TrackerDealer`.

`TrackerUD`: Standard staircase.
`TrackerBinom`: Keeps track of probability of null hypothesis that subject is guessing. Useful for training subjects to a certain p-value, though note the inherent multiple comparisons problem. Probably best to use p = 0.02 or so.
`TrackerDealer`: a group of trackers. It handles trial order so that all trackers stay within a certain number of trials or reversals, but randomizes as much as possible within that constraint.

Lots more documentation to come, but I wanted to get this in as a PR so my undergrad, Maddy, could take a look at it.

Feel free to make comments now if you can make heads or tails of it.

This code should do a 100 trial run that uses a weighted up-down (Kaernbach, 1991) to converge at 75% correct.

```python
import numpy as np
from expyfun.stimuli import TrackerUD
def callback(event_type, value=None, timestamp=None):
    print(event_type, value, timestamp)
tr = TrackerUD(callback, 1, 1, 0.06, 0.02, 100, 'trials', 1)
while not tr.stopped:
    tr.respond(np.random.rand() < tr.x_current)
tr.plot()
```


expyfun/_version.py
```diff
@@ -1,2 +1,2 @@
 __version__ = '2.0.0.dev0'
-__fork__ = '/Eric89GXL'
+__fork__ = 'Unknown'
```
Eric89GXL: whoops!

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,629 @@
...
+        callback = callback_dummy
+    elif isinstance(callback, ExperimentController):
+        callback = callback.write_data_line
+
+    if callable(callback):
+        return callback
+    else:
+        raise(TypeError,
+              'callback must be a callable, None, or an instance of '
+              'ExperimentController.')
```
Eric89GXL: This can be simplified a tiny bit to:
```
if not callable(callback):
    raise ...
return callback
```

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,629 @@
...
+            stop_rule=stop_rule,
+            start_value=start_value,
+            change_criteria=change_criteria,
+            change_rule=change_rule,
+            x_min=x_min,
+            x_max=x_max,
+            start_1_down=start_1_down)))
+
+    def respond(self, correct):
+        if not self._stopped:
```
Eric89GXL: you can save a ton of indented lines by doing `if self._stopped: raise...` then you don't need an `else` / indentation

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,629 @@
...
+                self._callback(
+                    'tracker_%i_stop' % self._tracker_id, json.dumps(dict(
+                    responses=list(self._responses.astype(int)),
+                    reversals=list(self._reversals),
+                    x=list(self._x))))
+        else:
+            raise(RuntimeError, 'Tracker is stopped.')
+
+    def _stop_here(self):
+        if str.lower(self._stop_rule) == 'reversals':
```
Eric89GXL: `if self._stop_rule.lower() == ...`

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,629 @@
...
+    # Display functions
+    # =========================================================================
+    def plot(self):
+        import matplotlib.pyplot as plt
+        line = plt.plot(1 + np.arange(self._n_trials), self._x, 'k.-')
+        dots = plt.plot(1 + np.where(self._reversals > 0)[0],
+                        self._x[self._reversals > 0], 'ro')
+        plt.show()
+        plt.xlabel('Trial number')
+        plt.ylabel('Level')
```
Eric89GXL: `plt` access is less good than something like:
```
fig, ax = plt.subplots(1)
ax.plot(...)
...
fig.tight_layout()
...
```

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,629 @@
...
+            return (self._tracker_history, self._x_history,
+                    self._response_history)
+        else:
+            inds = np.invert(np.isnan(self._response_history))
+            return (self._tracker_history[inds], self._x_history[inds],
+                    self._response_history[inds].astype(bool))
+
+
+# =============================================================================
+# Junk functions for testing (remove on release)
```
Eric89GXL: these should be moved to proper test functions instead of removed

expyfun/_version.py
rkmaddox: What should this be? (Never quite understood this functionality)

Eric89GXL: @drammock?

rkmaddox: @drammock?

drammock: it really should say `'LABSN'`.  If it says `'Unknown'` then `setup.py` did not detect a `.git` directory, meaning expyfun was probably downloaded as a zip file (or was installed via `download_version`).  If it says anything else, it reflects `git config --get remote.origin.url` which is a really bad idea, since we (the devs) have an unwritten convention of setting the LABSN remote as `upstream` and setting our personal forks as `origin`.  I move that you leave this as-is for now (as `'Unknown'`), and that the `__fork__` machinery be dismantled and removed.  My use case for it was marginally justifiable at the time, and no longer exists.

drammock: see #305

expyfun/stimuli/_tracker.py
rkmaddox: It has always seemed like the member functions of axes objects are spotty. For example, `ax.xlabel(...)` doesn't work. Is `ax.xaxis.set_label('Foo')` the right way to do it?

Eric89GXL: `ax.set(xlabel=...)`

rkmaddox: Got it, thanks. That's actually handier than having to do separate lines to set things!

rkmaddox:
Lots of changes, still not quite ready for review. No tests yet...

rkmaddox:
@maddycapp27 Feel free to comment on the documentation wherever things are unclear.

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,858 @@
...
+        The function that will be used to print the data, usually to the
+        experiment .tab file. It should follow the prototype of
+        ``ExperimentController.write_data_line``. If an instance of
+        ``ExperimentController`` is given, then it will take that object's
+        ``write_data_line`` function. If None is given, then it will not write
+        the data anywhere.
+    up : int
+        The number of wrong answers necessary to move the tracker level up.
+    down : int
+        The number of correct answers necessary to move the tracker level down.
```
drammock: FWIW, I find this aspect of the implementation a bit confusing.  I think of adaptive trackers as modulating "difficulty", with difficulty being implemented in different ways (noise level, separation angle, etc) depending on the experiment.  On that view, correct answers would move the difficulty **up**, not down.  Here, you're talking in terms of "tracker level", for which I don't have intuitions (and hence, might be more likely to make mistakes in coding an experiment).  Is there a convention I'm not aware of in which "tracker level" roughly equates to "how easy the task is", such that this implementation makes sense to everyone except me?

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,858 @@
...
+    step_size_up : float | list of float
+        The size the step when the tracker moves up. If float it will stay the
+        same. If list of float then it will change when when
```
drammock: "when when"

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,858 @@
...
+    tracker (staircase), or it can implement a weighted up-down procedure.
+
+    Parameters
+    ----------
+    callback : callable | ExperimentController | None
+        The function that will be used to print the data, usually to the
+        experiment .tab file. It should follow the prototype of
+        ``ExperimentController.write_data_line``. If an instance of
+        ``ExperimentController`` is given, then it will take that object's
+        ``write_data_line`` function. If None is given, then it will not write
```
drammock: ticks around None

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,858 @@
...
+        The number of wrong answers necessary to move the tracker level up.
+    down : int
+        The number of correct answers necessary to move the tracker level down.
+    step_size_up : float | list of float
+        The size the step when the tracker moves up. If float it will stay the
+        same. If list of float then it will change when when
+        ``change_criteria`` are encountered. See note below for more specific
+        information on dynamic tracker parameters specified with a list.
+    step_size_down : float | list of float
+        The size the step when the tracker moves down. If float it will stay
```
drammock: The size **of** the step

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,858 @@
...
+        the same. If list of float then it will change when when
```
drammock: another "when when"

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,858 @@
...
+        ``change_criteria`` are encountered. See note below for more specific
+        information on dynamic tracker parameters specified with a list.
+    stop_criterion : int
+        The number of reversals or trials that will stop the tracker.
+    stop_rule : str
+        How to determined when the tracker stops. Can be 'reversals' or
+        'trials.'
```
drammock: for clarity, period should be outside quotes here.

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,858 @@
...
+    first element of change_criteria must always be 0.
+    For the example given above:
+        ``step_size_up=[1., 0.2]``, ``step_size_down=[1., 0.2]``,
+        ``change_criteria=[0, 2]``, ``change_rule='reversals'``
+    would change the step sizes from 1 to 0.2 after two reversals.
+
+    If dynamic step sizes are used, both ``step_size_up`` and
+    ``step_size_down`` must have the same length as each other and
+    ``change_cirteria``. If static step sizes are used, both ``step_size_up``
+    and ``step_size_up`` must be scalars and ``change_criteria`` must be None.
```
drammock: first sentence of this paragraph is redundant with the final sentence of previous paragraph (and also has a misspelled "change_cirteria").  I recommend deleting first sentence of this one.

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,858 @@
...
+# =============================================================================
+# Set up the logging callback (use write_data_line or do nothing)
+# =============================================================================
+def callback_dummy(event_type, value=None, timestamp=None):
+    """Take the arguments of write_data_line, but do nothing.
+    """
+    pass
+
+
+def check_callback(callback):
```
drammock: (same question re: `callback_dummy`)

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,858 @@
...
+        self._stop_criterion = stop_criterion
+        self._stop_rule = stop_rule
+        self._start_value = start_value
+        self._x_min = x_min
+        self._x_max = x_max
+
+        if change_criteria is None:
+            change_criteria = [0]
+        elif change_criteria[0] != 0:
+            raise(ValueError('First element of change points must be 0.'))
```
drammock: Better error message: `First element of change_criteria must be 0.`

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,858 @@
...
+        self._change_criteria = np.asarray(change_criteria)
+        if change_rule not in ['trials', 'reversals']:
+            raise(ValueError, "change_rule must be either 'trials' or "
+                  "'reversals'")
+        else:
+            self._change_rule = change_rule
```
drammock: this need not be within an `else` block

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,858 @@
...
+
+        if np.isscalar(step_size_up):
+            step_size_up = [step_size_up] * len(change_criteria)
```
drammock: would `step_size_up = np.atleast_1d(step_size_up)` do this job in one line, without the `if` clause?  You're already testing below for equivalent lengths...

expyfun/stimuli/_tracker.py
rkmaddox: The level would equate to something like intensity or contrast. This is in line with the "good thresholds are low" frame of reference, and also results in a 3-down 1-up tracker operating as they are described in the literature.

expyfun/stimuli/_tracker.py
rkmaddox: They aren't imported, but I'll make them private anyway.

expyfun/stimuli/_tracker.py
rkmaddox: Not the same, because this will turn a scalar `step_size_up` into a list-like of the proper length. However, I don't like that functionality--it allows too much sloppiness. So I will change to what you suggest and require either ALL scalars or ALL list-like of the proper length.

rkmaddox:
Thanks @drammock for the fine-combed read-through.

drammock:
(there may be more coming; I had to abort midway through to go to a meeting, will resume this afternoon)

rkmaddox:
OK, I believe all comments are addressed. Coverage of new stuff is north of 99%. Tests pass locally and only CI failures are unrelated to the tracker code. Should be ready for final review.

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
+"""Adative tracks for psychophysics (individual, or multiple randomly dealt)
```
maddycapp27: Adaptive

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+    return callback
+
+
+# =============================================================================
+# Define the TrackerUD Class
+# =============================================================================
+class TrackerUD(object):
+    """Up-down adaptive tracker
+
+    This class implements a standard up-down adative tracker object. Based on
```
maddycapp27: adaptive

expyfun/stimuli/_tracker.py
maddycapp27: Adaptive

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+        dynamic tracker parameters specified with a list.
+    step_size_down : float | list of float
+        The size of the step when the tracker moves down. If float it will stay
+        the same. If list of float then it will change when ``change_criteria``
+        are encountered. See note below for more specific information on
+        dynamic tracker parameters specified with a list.
+    stop_criterion : int
+        The number of reversals or trials that will stop the tracker.
+    stop_rule : str
+        How to determined when the tracker stops. Can be 'reversals' or
```
maddycapp27: determine

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+        The starting level of the tracker.
+    change_criteria : list of int | None
+        The points along the tracker to change its step sizes. Has an effect
+        where ``step_size_up`` and ``step_size_down`` are
+        lists. The length of ``change_criteria`` must be the same as the
+        length of ``step_size_up`` and ``step_size_down``. See note below for
+        more specific usage information. Should be None if static step sizes
+        are used.
+    change_rule : str
+        Whether to change paramters based on 'trials' or 'reversals'.
```
maddycapp27: parameters

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+        self._stop_criterion = stop_criterion
+        self._stop_rule = stop_rule
+        self._start_value = start_value
+        self._x_min = x_min
+        self._x_max = x_max
+
+        if change_criteria is None:
+            change_criteria = [0]
+        elif change_criteria[0] != 0:
+            raise(ValueError('First element of change_crieria must be 0.'))
```
maddycapp27: change_criteria

expyfun/stimuli/tests/test_tracker.py
```diff
@@ -0,0 +1,154 @@
+import numpy as np
+from expyfun.stimuli import TrackerUD, TrackerBinom, TrackerDealer
+from expyfun import ExperimentController
+from nose.tools import assert_raises  # , assert_equal, assert_true
```
drammock: reminder to remove unused import comment

expyfun/stimuli/tests/test_tracker.py
```diff
@@ -0,0 +1,154 @@
...
+    assert_raises(ValueError, TrackerUD, None, 3, 1, [1, 0.5], [1, 0.5], 10,
+                  'trials', 1, change_criteria=[0])
+    assert_raises(ValueError, TrackerUD, None, 3, 1, [1], [1, 0.5], 10,
+                  'trials', 1, change_criteria=[0, 2])
+    assert_raises(ValueError, TrackerUD, None, 3, 1, [1, 0.5], [1], 10,
+                  'trials', 1, change_criteria=[0, 2])
+    assert_raises(ValueError, TrackerUD, None, 3, 1, [1, 0.5], [1, 0.5], 10,
+                  'trials', 1, change_criteria=[0, 2], change_rule='foo')
+    assert_raises(ValueError, TrackerUD, None, 3, 1, [1, 0.5], [1, 0.5], 10,
+                  'trials', 1)
```
drammock: it can be helpful to have comments explaining which conditions are being violated in each call to `assert_raises` (e.g., a comment like `# test mismatching lengths of step_size_up and change_criterion`)

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
+"""Adaptive tracks for psychophysics (individual, or multiple randomly dealt)
+"""
+# Author: Ross Maddox <ross.maddox@rochester.edu>
+#
+# License: BSD (3-clause)
+
+import numpy as np
+import time
+from scipy.stats import binom as binom
```
drammock: don't need `as binom` since you're not changing its name

expyfun/stimuli/_tracker.py
drammock: fair enough.

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+    encountered. This means that the length of the list defining a dynamic
+    parameter must always be the same as that of ``change_criteria``, and the
+    first element of change_criteria must always be 0.
+    For the example given above:
+        ``step_size_up=[1., 0.2]``, ``step_size_down=[1., 0.2]``,
+        ``change_criteria=[0, 2]``, ``change_rule='reversals'``
+    would change the step sizes from 1 to 0.2 after two reversals.
+
+    If static step sizes are used, both ``step_size_up``
+    and ``step_size_up`` must be scalars and ``change_criteria`` must be None.
```
drammock: one of these should be `step_size_down` (prob. the first one, to avoid extra line break)

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+        multiple comparisons issues and a more stringent ``alpha`` should be
+        considered.
+    chance : float
+        The chance level of the task being performed. Must be between 0 and 1.
+    max_trials : int
+        The maximum number of trials to run before stopping the track without
+        reaching ``alpha``.
+    min_trials : float | list of float
+        The minimum number of trials to run before allowing the track to stop
+        on reaching ``alpha``. Has no effect if if ``stop_early`` is ``False``.
```
maddycapp27: one if

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+            change_criteria = [0]
+        elif change_criteria[0] != 0:
+            raise(ValueError('First element of change_criteria must be 0.'))
+        self._change_criteria = np.asarray(change_criteria)
+        if change_rule not in ['trials', 'reversals']:
+            raise(ValueError, "change_rule must be either 'trials' or "
+                  "'reversals'")
+        self._change_rule = change_rule
+
+        step_size_up = list(np.atleast_1d(step_size_up))
```
drammock: wondering why you're casting to a list, when a few lines later you cast back to array.  Also, want to call your attention to what might happen if the user gives a 2D array for `step_size_up`...  as currently implemented, I think it will still pass these tests as long as its first dimension is the right length.  Presumably it will error out later, but might be worth catching it here?

drammock: Example:
```
step_size_up = np.atleast_1d(step_size_up)
if step_size_up.ndim != 1:
    raise ValueError('step_size_up has too many dimensions (must be scalar or 1-D)')
```

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+
+    @property
+    def reversal_inds(self):
+        """The trial indices which had reversals"""
+        return np.where(self._reversals)[0]
+
+    # =========================================================================
+    # Display functions
+    # =========================================================================
+    def plot(self):
```
drammock: ideally would like this to have an argument `ax=None` so it could plot on an existing axes object if provided.  Could be a later PR if you're sick of this one though.  :sweat_smile:

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+        self._trial_complete = True
+        self._response_history = np.append(self._response_history, correct)
+
+    def history(self, include_skips=False):
+        """The history of the dealt trials and the responses
+
+        Inputs
+        ------
+            include_skips : boolean
+                Whether or not to include trails where a tracker was dealt but
```
maddycapp27: trials

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+
+        Returns
+        -------
+        line : list Line2D
+            The handle to the threshold line, as returned from ``plt.plot``.
+        """
+        if ax is None:
+            ax = plt.gca()
+        h = ax.plot([1, self._n_trials], [self.threshold(n_skip)] * 2,
+                    '--', color='gray')
```
drammock: this might be a good place for `**kwargs` rather than hard-coding the linetype and color.

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+        return h
+
+    def threshold(self, n_skip=2):
+        """Plot a line showing the threshold on the current axes.
```
drammock: wrong docstring, this computes the threshold, not plots it

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+    callback : callable | ExperimentController | None
+        The function that will be used to print the data, usually to the
+        experiment .tab file. It should follow the prototype of
+        ``ExperimentController.write_data_line``. If an instance of
+        ``ExperimentController`` is given, then it will take that object's
+        ``write_data_line`` function. If ``None`` is given, then it will not
+        write the data anywhere.
+    alpha : float
+        The p-value which is considered significant. Must be between 0 and 1.
+        Note that if ``stop_early`` is `true` there is the potential for
```
drammock: capitalize `True`

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+        return 'trials'
+
+
+# =============================================================================
+# Define a container for interleaving several tracks simultaneously
+# =============================================================================
+# TODO: Make it so you can add a list of values for each dimension (such as the
+# phase in a BMLD task) and have it return that
+
+# TODO: eventually, make a BaseTracker class so that TrackerDealer can make
```
drammock: TODOs should be migrated to repo issues?  WDYT @Eric89GXL?  Are we OK with TODOs in the source?

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+    shape : list-like
+        The dimensions of the tracker container.
+    tracker_class : class
+        The class for the tracker you want to run, e.g. ``TrackerUD``--not an
+        instance.
+    tracker_args : list-like
+        The arguments used to instantiate each of the trackers.
+    tracker_kwargs : dict
+        The keyword arguments used to instantiate each of the trackers.
+    slop : int
```
drammock: is `max_lag` a more descriptive name here?

drammock:
just sent a bunch more comments.  I haven't carefully vetted the math, but I did read through it and didn't notice anything obviously wrong.  :+1: for merge when existing comments addressed and CIs are happy.

Eric89GXL:
Let me know when @drammock's comments are addressed and the CIs are happy and I'll have a look

rkmaddox:
Will do. Won't get to it today and tomorrow need to make a lecture. Shouldn't take too long though, so I'l try to sneak it in.

expyfun/stimuli/_tracker.py
rkmaddox: I first just did it as a ``numpy`` array, but then the json writer from ``write_data_line`` couldn't handle it. It can handle lists, though.

expyfun/stimuli/_tracker.py
rkmaddox: It was quick. Done.

expyfun/stimuli/_tracker.py
drammock: Oh yeah, I've hit that limitation of JSON before.  In that case, maybe the casting to list belongs in the call to `self._callback` rather than here?

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,837 @@
...
+        self._tracker_id = id(self)
+        self._callback('tracker_identify', json.dumps(dict(
+            tracker_id=self._tracker_id,
+            tracker_type='TrackerUD')))
+
+        self._callback('tracker_%i_init' % self._tracker_id, json.dumps(dict(
+            callback=None,
+            up=up,
+            down=down,
+            step_size_up=step_size_up,
```
drammock: ...i.e., this line would become `step_size_up = list(step_size_up)`, or `step_size_up = step_size_up.tolist()` depending on which way you swing.

rkmaddox: Ah, sure that's better. I was thinking I would have to modify the callback itself, but of course that's not true.

expyfun/stimuli/_tracker.py
rkmaddox: Done (sorry about all the "dones" you are about to receive, I'm just trying to keep track)

expyfun/stimuli/_tracker.py
rkmaddox: Done.

expyfun/stimuli/_tracker.py
rkmaddox: Done.

expyfun/stimuli/_tracker.py
rkmaddox: Done.

expyfun/stimuli/tests/test_tracker.py
rkmaddox: Done.

expyfun/stimuli/_tracker.py
rkmaddox: Done.

expyfun/stimuli/_tracker.py
rkmaddox: Yes, and changed, but it does lead to this lovely little snippet: ``if lag_max > self._max_lag``.

Worth it, though, I think (as much as I would like to have a parameter called "slop").

expyfun/stimuli/_tracker.py
rkmaddox: What sayest thou, @Eric89GXL? I should be totally honest: these aren't things I'm likely to address in the near future, so the issues will likely sit there a while.

expyfun/stimuli/_tracker.py
rkmaddox: I started doing this, but there's two plotting commands (lines for staircase and dots for reversals), which makes it weird. This is meant to be a quick and dirty plot, probably not so much for making publications. If people want to make paper-worthy plots the data they need are public.

expyfun/stimuli/_tracker.py
rkmaddox: I am sick of the PR, though.

expyfun/stimuli/_tracker.py
rkmaddox: Done.

expyfun/stimuli/_tracker.py
drammock: in that case I'd rather have TODOs in the source than lingering issues.  I'm guilty of several of those already; the fewer the better.

expyfun/stimuli/tests/test_tracker.py
rkmaddox: Done.

expyfun/stimuli/_tracker.py
drammock: that is true of the `plot()` method but not the `plot_thresh()` method.  But you raise a good point that it would be hard to do cleanly for the `plot()` method, so for consistency I'll agree to leave both methods as-is (with hard-coded styles).

rkmaddox:
OK, I am stuck. The python 3.5 builds in AppVeyor and Travis fail because of a JSON thing. I have been trying to track this down, but I can't figure it out.

Then the OSX build in Travis is also failing on my tests, but it looks like it's an ExperimentController thing. I am confused.

None of these failures are happening on my machine, so I am having trouble figuring out how to fix them.

Eric89GXL:
Are you running Anaconda? If so you should be able to easily set up a Py3k test env. If not, I can try it.

Eric89GXL:
You need the elements in your lists to be `int` not `np.int64`:
```
(py35) larsoner@bunk:~/Documents/python/expyfun$ nosetests expyfun/stimuli/tests/test_tracker.py:test_tracker_dealer --pdb
Test TrackerDealer ... > /home/larsoner/miniconda/envs/py35/lib/python3.5/json/encoder.py(179)default()
-> raise TypeError(repr(o) + " is not JSON serializable")
(Pdb) p o
0
(Pdb) p type(o)
<class 'numpy.int64'>
(Pdb) u
> /home/larsoner/miniconda/envs/py35/lib/python3.5/json/encoder.py(256)iterencode()
-> return _iterencode(o, 0)
(Pdb) p o
{'step_size_down': [0.02], 'x_max': inf, 'up': 1, 'stop_criterion': 20, 'change_criteria': [0], 'down': 1, 'start_value': 1, 'stop_rule': 'reversals', 'step_size_up': [0.059999999999999998], 'x_min': -inf, 'change_rule': 'reversals', 'callback': None}
(Pdb) p type(o['change_criteria'])
<class 'list'>
(Pdb) p type(o['change_criteria'][0])
<class 'numpy.int64'>
```

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,857 @@
...
+    def __iter__(self):
+        self._index = 0
+        return self
+
+    def next(self):
+        if self._index == len(self._trackers):
+            raise(StopIteration)
+        t = self._trackers[self._index]
+        self._index += 1
+        return t
```
Eric89GXL: I think for Py3k you also need to provide `__next__`

rkmaddox: It would look like this, yeah?

```python
def __next__(self):
    return self.next()
```

Eric89GXL: yeah that should work

Eric89GXL:
Unorderable types error

https://travis-ci.org/LABSN/expyfun/jobs/218914507#L920

expyfun/stimuli/tests/test_tracker.py
```diff
@@ -88,11 +89,17 @@ def test_tracker_ud():
...
                   'trials', 1)
 
+    # start_value scalar type checking
+    assert_raises(TypeError, TrackerUD, None, 3, 1, [1, 0.5], [1, 0.5], 10,
+                  'trials', [9, 5], change_criteria=[0, 2])
+    assert_raises(TypeError, TrackerUD, None, 3, 1, [1, 0.5], [1, 0.5], 10,
+                  'trials', None, change_criteria=[0, 2])
+
 
 def test_tacker_binom():
```
drammock: just noticed this typo: should be `test_tracker_binom`

rkmaddox:
Other tests are running, but I do not know how to fix the sphinx matplotlib backend stuff from CircleCI.

expyfun/stimuli/tests/test_tracker.py
rkmaddox: Whoops. I'm gonna let these tests finish and then I will fix and push.

expyfun/stimuli/tests/test_tracker.py
```diff
@@ -0,0 +1,184 @@
+import numpy as np
+from expyfun.stimuli import TrackerUD, TrackerBinom, TrackerDealer
+from expyfun import ExperimentController
+from nose.tools import assert_raises
+import matplotlib.pyplot as plt
```
drammock: I think you can fix the CircleCI error by inserting `import matplotlib; matplotlib.use('Agg')` _above_ this line that imports `pyplot`.  See [here](https://github.com/LABSN/expyfun/blob/master/expyfun/analyze/tests/test_viz.py#L11)

drammock: (note that in `test_viz.py` we import `pyplot` only within the test(s) that need it, not globally)

rkmaddox: That does not fix it locally. I messed around with ``conf.py`` (following the error message's suggestions) and that did not fix it either.

rkmaddox:
AppVeyor finally happy, Travis at 23 minutes and counting for its last test...

rkmaddox:
Travis failed: https://travis-ci.org/LABSN/expyfun/jobs/221053005#L550

Looks like it's a problem with creating the audio controller? I took that stuff directly from ``test_viz.py`` and it seems to work fine there. Any ideas what I am doing wrong? I've spent almost as much time on these last couple errors as I did coding the damn thing.

expyfun/stimuli/tests/test_tracker.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+std_kwargs = dict(output_dir=None, full_screen=False, window_size=(1, 1),
+                  participant='foo', session='01', stim_db=0.0, noise_db=0.0,
+                  trigger_controller='dummy', response_device='keyboard',
+                  audio_controller='pyglet',
+                  verbose=True, version='dev')
+
+
+def test_tracker_ud():
```
Eric89GXL: as with the `ExperimentController` tests, you should use these decorators:
```
@_hide_window
@requires_opengl21
```

expyfun/stimuli/tests/test_tracker.py
```diff
@@ -0,0 +1,189 @@
...
+                  'trials', 1)
+
+    # start_value scalar type checking
+    assert_raises(TypeError, TrackerUD, None, 3, 1, [1, 0.5], [1, 0.5], 10,
+                  'trials', [9, 5], change_criteria=[0, 2])
+    assert_raises(TypeError, TrackerUD, None, 3, 1, [1, 0.5], [1, 0.5], 10,
+                  'trials', None, change_criteria=[0, 2])
+
+
+def test_tracker_binom():
```
Eric89GXL: same here

expyfun/stimuli/_tracker.py
```diff
@@ -0,0 +1,862 @@
...
+"""
+# Author: Ross Maddox <ross.maddox@rochester.edu>
+#
+# License: BSD (3-clause)
+
+import numpy as np
+import time
+from scipy.stats import binom
+import json
+import matplotlib.pyplot as plt
```
Eric89GXL: nest this import where you use `plt` and it will fix CircleCI

doc/conf.py
```diff
@@ -12,13 +12,13 @@
...
 
+import sphinx_gallery  # noqa
+import sphinx_bootstrap_theme
 import inspect
 import os
 from os.path import relpath, dirname
 import sys
 from datetime import date
-import sphinx_gallery  # noqa
-import sphinx_bootstrap_theme
```
Eric89GXL: why you messin with my import order?

The idea was to go system-level first, then more specific later...

drammock:
hmm... [pyglet could not be imported](https://travis-ci.org/LABSN/expyfun/jobs/221053005#L469) which happens [here](https://github.com/LABSN/expyfun/blob/master/expyfun/_sound_controllers.py#L26-L34).  Is that handled by the openGL decorator?

doc/conf.py
rkmaddox: The error message said to do all ``'sphinx`` related imports before ``matplotlib`` imports in ``conf.py``. It didn't fix it, though.

Eric89GXL:
that or hide_window, can't remember

Eric89GXL:
It's also possible we have some other underlying bug we need to fix, but the critical thing for @rkmaddox is that those decorators will make his tests pass :)

expyfun/stimuli/_tracker.py
rkmaddox: I did that in a commit that is still on my local system and it doesn't fix it. I'll push it so you can see.

doc/conf.py
Eric89GXL: but there is no `matplotlib` import here, so you're barking up the wrong tree

expyfun/stimuli/tests/test_tracker.py
Eric89GXL: ... these will "fix" Travis

expyfun/stimuli/_tracker.py
rkmaddox: When I do ``make html_dev`` locally I still get the same error as CircleCI is getting when trying to build the documentation.

drammock:
@Eric89GXL I'm skeptical that this will fix it, since [the thing travis is choking on](https://travis-ci.org/LABSN/expyfun/jobs/221062279#L542-L548) is the pyglet `SoundPlayer` and `AudioFormat` objects, not the window.

Eric89GXL:
@drammock that is what I mean about "some other underlying bug". It won't fix it, but it should work around the failures at least, as the other tests that use that decorator don't fail (though they may be skipped).

Eric89GXL:
CircleCI does still seem unhappy. Do the docs build for you on `master` @rkmaddox?

expyfun/stimuli/_tracker.py
Eric89GXL: Wait, this still hasn't been moved...?

expyfun/stimuli/_tracker.py
Eric89GXL: I'm fine with TODOs in the source for minor, non-show-stopping things that are primarily code-design issues (like these).

Eric89GXL:
Decorators
-----------
@drammock I remembered why those decorators work the way they do. The failure is on Travis OSX where sound can't be played. Sounds are only played when `ExperimentController` is used. Those decorators should be used anytime `ExperimentController` is used in order to (among other things) to hide windows when running tests. So even though the error message is a bit cryptic, the CI failure does at least indicate something meaningful/useful. Sorry it was a bit of a wild goose chase here @rkmaddox... maybe we can think of a way to make it clearer, or add a note to `.travis.yml` or something. Where did you look to resolve the error?

Fixes
-----
I've pushed a commit that:

1. nests the `matplotlib` import
2. moves two decorators to the correct function (they had decorated the wrong function, thus the continued Travis failure)
3. fixes a couple of small style issues with your latest commit
4. adds your classes to `python_reference.rst`
5. fixes a small docstring error
6. adds a check to `TrackerDealer`.

With those changes, Circle is now happy, and I expect Travis will come back happy, too:

https://148-11614571-gh.circle-artifacts.com/0/home/ubuntu/expyfun/doc/_build/html/python_reference.html#stimulus-design

@drammock are you happy with the changes?

Remaining work
---------
1. There are no updated examples. It would be nice to have one to at least show how some basic functionality works. @rkmaddox do you have time to do it? You should use `building_doc` to skip any interactive steps for automated doc building / CircleCI (feel free to `git grep` to see how it's used elsewhere) but still give users who actually execute the script the right experiences.

2. It seems like the TrackerDealer could be simplified by having the user instantiate their own instances. Then they can set their own params, and you can get rid of the comment " TODO: Make it so you can add a list of values for each dimension". To replicate existing behavior users just do:
    ```
    trackers = [TrackerUD(*tracker_args, **tracker_kwargs) for _ in range(prod(shape))]
    ```
    Which is easy enough, and will actually become clearer/more explicit just by passing the args and kwargs as normal instead of as list and dict. And if they want different parameters, they can explicitly set them. This seems like a more future-compatible design. Within the class, I think we just need to check that they have compatible parameters, whatever those are (number of trial or something?). WDYT?

drammock:
The API changes look fine to me, as long as @rkmaddox is happy with them.

Eric89GXL:
Updated example:

https://156-11614571-gh.circle-artifacts.com/0/home/ubuntu/expyfun/doc/_build/html/auto_examples/stimuli/plot_tracker_staircase.html

I'll go ahead and merge once the CIs come back happy.

Thanks for working through it @rkmaddox @drammock

rkmaddox:
Awesome, thanks for the help, all!

PR #303, opened 2017-03-23
MRG: Dont mangle env during setup

### LINKED ISSUE #271: Error on download_version
We are trying to download a legacy version of `expyfun` but getting the following error:

``` python
>>> expyfun.download_version('090948e')
  File "<stdin>", line 1, in <module>
  File "c:\users\labuser\builds\expyfun\expyfun\_git.py", line 94, in download_version
    assert git_version()[0].lower() == version[:7].lower()
AssertionError
```

When I do `pdb` I get that `git_version()` is `'Unknown'`. Any ideas?

### LINKED ISSUE #305: dismantle __fork__ tracking in setup.py
this was a slightly bad idea at the time, and is no longer necessary.

Eric89GXL:
@drammock on OSX setting the `env` during `setup.py` can break things because it makes `git` no longer available. I think you set at least some of this up -- think it's reasonable?

Closes #271.
Closes #305.

### DIFF
--- a/expyfun/tests/test_version.py
+++ b/expyfun/tests/test_version.py
@@ -25,21 +25,18 @@ def test_version():
     assert_true(all('actual' in str(ww.message) for ww in w))
 
     v = 'b798e65'
-    f = 'drammock'
     if not _has_git:
-        assert_raises(ImportError, download_version, v, tempdir, f)
+        assert_raises(ImportError, download_version, v, tempdir)
     else:
-        assert_raises(IOError, download_version, v, op.join(tempdir, 'foo'), f)
-        assert_raises(RuntimeError, download_version, 'x' * 7, tempdir, f)
-        download_version(v, tempdir, f)
+        assert_raises(IOError, download_version, v, op.join(tempdir, 'foo'))
+        assert_raises(RuntimeError, download_version, 'x' * 7, tempdir)
+        download_version(v, tempdir)
         ex_dir = op.join(tempdir, 'expyfun')
         assert_true(op.isdir(ex_dir))
         assert_true(op.isfile(op.join(ex_dir, '__init__.py')))
         with open(op.join(ex_dir, '_version.py')) as fid:
             line1 = fid.readline().strip()
-            line2 = fid.readline().strip()
         assert_equal(line1.split(' = ')[1][-8:-1], v)
-        assert_equal(line2.split(' = ')[1].strip('\''), f)
 
         # auto dir determination
         orig_dir = os.getcwd()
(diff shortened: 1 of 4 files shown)


drammock:
Looks reasonable at first glance. Did you do a `git grep Unknown`? We might test for `fork='Unknown'` somewhere... The download_version code is so hard to test automatically, I wish we had a cleaner way of handling it.

Eric89GXL:
> We might test for fork='Unknown' somewhere...

How does that affect this PR? I don't think I changed that behavior.

drammock:
It looked from the diff like `fork` is no longer getting set as unknown, so that's why I mentioned it. Sorry, I'm probably misunderstanding something... Traveling and viewing on phone makes for bad code reviews.

Eric89GXL:
I just set those variables at the top as the fallback. Actually there were branches in the code where it wasn't set at all, which led to missing variable errors in some cases.

drammock:
Oh, I see that now. I'll be out of town for a week, so if this is urgent you should merge it yourself when it's ready.

Eric89GXL:
Nope, feel free to come back to it later

Eric89GXL:
@drammock should I just get rid of the FORK stuff in this PR?

drammock:
sure, if you're up for it, feel free to kill the fork-detection code here

expyfun/tests/test_version.py
```diff
@@ -25,21 +25,18 @@ def test_version():
     assert_true(all('actual' in str(ww.message) for ww in w))
 
     v = 'b798e65'
```
drammock: If you change this to a commit hash that happens to start with 0, it reveals a bug on [this line](https://github.com/LABSN/expyfun/pull/303/files#diff-2175524ba23bd6005be929c15c98478eR86)... need to remove the `[0]` in the LHS.  Noticed this by switching `v` here to the commit 090948e that led to #271

setup.py
```diff
@@ -31,45 +31,23 @@ def git_version():
...
         # minimal env; LANGUAGE is used on win32
-        env = dict(LANGUAGE='C', LANG='C', LC_ALL='C')
-        for k in ['SYSTEMROOT', 'PATH']:
-            v = os.environ.get(k)
-            if v is not None:
-                env[k] = v
         return subprocess.Popen(cmd, stdout=subprocess.PIPE,
-                                env=env).communicate()[0]
+                                ).communicate()[0]
+    GIT_REVISION = "Unknown"
```
drammock: single quotes.

Eric89GXL:
Comments addressed and CIs are happy

drammock:
thanks @Eric89GXL

PR #306, opened 2017-04-05
MRG: hrtf interpolation

maddycapp27:
interp_hrtf is a function for interpolating HRTFs between known step sizes based on averaging amplitude and phase in the frequency domain based on this paper: http://www.dtic.mil/dtic/tr/fulltext/u2/a470084.pdf  The function will take the desired angle, compute the two nearest angles for which there is a known HRTF, and then calculate an interpolated HRTF.

Here is an example of the code interpolating an HRTF at 5 degrees using known HRTFs at 0 and 10 degrees. Each ear is shown on a different subplot.
5degcompare

Hopefully this can allow convolve_hrtf to pull brir's from the interp_hrtf function if the input angle is not from a set of experimentally measured HRTFs.

I still need to generalize the source so it doesn't just take cipic. The only things to change are in lines 181 and 182 when convolve_hrtf is called and line 176 where the interval between the nearest known HRTFs is set--not sure how best to implement the difference in interval.


expyfun/stimuli/_hrtf.py
```diff
@@ -151,3 +151,67 @@ def convolve_hrtf(data, fs, angle, source='barb'):
         brir = [resample(b, fs, brir_fs) for b in brir]
     out = np.array([np.convolve(data, brir[o]) for o in order])
     return out
+
+    
+def make_sym(x):
```
Eric89GXL: prefix with `_` to make it clear it's private (not for public use)

expyfun/stimuli/_hrtf.py
```diff
@@ -151,3 +151,67 @@ def convolve_hrtf(data, fs, angle, source='barb'):
...
+    forces symmetry around zero
+    """
+    n = x.shape[-1]
+    x[..., :-((n + 1) // 2):-1] = np.conj(x[..., 1:(n + 1) // 2])
+    if np.mod(n, 2) == 0:
+        x[..., n // 2] = 0
+    return x
+    
+
+def interp_hrtf(angle, fs):
```
Eric89GXL: I don't think we need a separate function for this. We can just add a parameter `interp=False` (by default) to our existing HRTF function. A user just does `interp=True` to explicitly say they are willing to use interpolation, then your code kicks in if the requested angle is not present.

expyfun/stimuli/_hrtf.py
```diff
@@ -151,3 +151,67 @@ def convolve_hrtf(data, fs, angle, source='barb'):
...
+    ----------
+    angle : float
+        The desired angle of the interpolated BRIR in degrees. The angle must
+        be between -90 and 90 degrees.
+    fs : int
+        The sampling frequency of the BRIR
+
+    Returns
+    -------
+    brir : 2 x 184 matrix containing interpolated BRIR for the given angle
```
Eric89GXL: please make this conform to the NumPy docstring standard (look at other expyfun functions for examples)

Also, please add a `References` section with the referenced paper

expyfun/stimuli/_hrtf.py
maddycapp27: Do you mean in convolve_hrtf (in which case there would be issues since my code calls convolve_hrtf to get the knowns) or in _get_hrtf?

Eric89GXL: Probably in `_get_hrtf`. It also has the list of angles, so you shouldn't need to call `convolve_hrtf` from within your function (this would be bad software design anyway)

rkmaddox:
Also, I took a look at the paper and it looks like they are averaging the log magnitude. In other words, you should be doing the geometric mean, not the arithmetic mean.

expyfun/stimuli/_hrtf.py
```diff
@@ -40,6 +40,19 @@ def _get_hrtf(angle, source, fs):
     0-degree BRIRs (mean of that across channels) is equal to 1. This will
     ensure that the RMS of a white signal filtered with this signal is
     unchanged.
+    
+    If interp is True, the function takes the two nearest known HRTFs (in the 
+    form of a BRIR) and uses them to calculate an HRTF (again, as a BRIR) 
+    between them for any given azimuthal angle between 0 and 90 degrees.
+    
+    Interpolation relies on averagin log magnitude and phase in the frequency
```
rkmaddox: averaginG

expyfun/stimuli/_hrtf.py
```diff
@@ -40,6 +40,19 @@ def _get_hrtf(angle, source, fs):
     0-degree BRIRs (mean of that across channels) is equal to 1. This will
     ensure that the RMS of a white signal filtered with this signal is
     unchanged.
+    
+    If interp is True, the function takes the two nearest known HRTFs (in the 
+    form of a BRIR) and uses them to calculate an HRTF (again, as a BRIR)
```
rkmaddox: Techincally, you're doing it in the frequency domain, so some of these caveats about BRIR are probably not necessary.

expyfun/stimuli/_hrtf.py
```diff
@@ -49,20 +62,56 @@ def _get_hrtf(angle, source, fs):
...
+        # combine magnitude and phase components
+        HRTF_a = np.multiply(HRTF_a_mag, np.exp(1j * (HRTF_a_phase)))
+        HRTF_a = _make_sym(HRTF_a)
+        brir_a = np.real(np.fft.ifft(HRTF_a))
+        brir = np.concatenate((brir_a[:, -delay:], brir_a[:, :-delay]), 1)
     return brir, data['fs'], leftward
 
 
-def convolve_hrtf(data, fs, angle, source='barb'):
+def convolve_hrtf(data, fs, angle, source='barb', interp=False):
```
rkmaddox: I think this would be a good time to change the default source to ``'cipic'``. Experiments that used ``'barb'`` are looong done, and CIPIC is a much better database. @Eric89GXL, @drammock? The only reason ``'barb'`` is included at all is for legacy, but it seems silly to default to the inferior database.

expyfun/stimuli/_hrtf.py
```diff
@@ -49,20 +62,56 @@ def _get_hrtf(angle, source, fs):
...
+        c = angles[idx[-1] + 1]
+        # get known brirs
+        brir_b =  brir[idx[-1]].copy()
+        brir_c = brir[idx[-1] + 1].copy()
+    
+        # find location of maximum for each component of each known brir and
+        # average to find a generally good place to declare time = 0 (all data
+        # from before this point gets shifted to the end)
+        delay = int(np.mean([np.argmax(brir_b, 1), np.argmax(brir_c, 1)]))
+        brir_b = np.concatenate((brir_b[:, delay:], brir_b[:, :delay]), 1)
```
rkmaddox: ``brir_b`` and ``brir_c`` as well as ``b`` and ``c`` and all other variables ending with those suffixes should be elements in a single array or list.

expyfun/stimuli/_hrtf.py
```diff
@@ -49,20 +62,56 @@ def _get_hrtf(angle, source, fs):
...
-    brir = brir[idx[0]].copy()
+    if not interp:
+        idx = np.where(angles == read_angle)[0]
+        assert len(idx) == 1
+        brir = brir[idx[0]].copy()
+    elif interp:
+        idx = np.where(angles < read_angle)[0]
+        # angles are
+        b = angles[idx[-1]]
+        c = angles[idx[-1] + 1]
```
rkmaddox: This could bomb if ``idx == len(angles)`` right?

expyfun/stimuli/_hrtf.py
```diff
@@ -49,20 +62,56 @@ def _get_hrtf(angle, source, fs):
...
+        # from before this point gets shifted to the end)
+        delay = int(np.mean([np.argmax(brir_b, 1), np.argmax(brir_c, 1)]))
+        brir_b = np.concatenate((brir_b[:, delay:], brir_b[:, :delay]), 1)
+        brir_c = np.concatenate((brir_c[:, delay:], brir_c[:, :delay]), 1)
+        # convert to frequency domain representation
+        HRTF_b = np.fft.fft(brir_b)
+        HRTF_c = np.fft.fft(brir_c)
+
+        # weighted averages of log magnitude and unwrapped phase
+        step = c-b
```
rkmaddox: space around operator

expyfun/stimuli/_hrtf.py
```diff
@@ -49,20 +62,56 @@ def _get_hrtf(angle, source, fs):
...
+        a = float(angle)
+        weight_b = (step - np.abs(a - b)) / step
+        weight_c = (step - np.abs(a - c)) / step
+        HRTF_a_logmag = (weight_b * np.log10(np.abs(HRTF_b)) +
```
rkmaddox: I think it should be computationally faster do do the geometric mean rather than the mean of the logs.

So: ``HRTF_a_mag = np.abs(HRTF_b) ** weight_b * np.abs(HRTF_c) ** weight_c``

expyfun/stimuli/_hrtf.py
drammock: as long as we can still call `'barb'` at need, I have no objection to changing the default

expyfun/stimuli/_hrtf.py
```diff
@@ -49,20 +62,56 @@ def _get_hrtf(angle, source, fs):
...
+        brir_c = brir[idx[-1] + 1].copy()
+    
+        # find location of maximum for each component of each known brir and
+        # average to find a generally good place to declare time = 0 (all data
+        # from before this point gets shifted to the end)
+        delay = int(np.mean([np.argmax(brir_b, 1), np.argmax(brir_c, 1)]))
+        brir_b = np.concatenate((brir_b[:, delay:], brir_b[:, :delay]), 1)
+        brir_c = np.concatenate((brir_c[:, delay:], brir_c[:, :delay]), 1)
+        # convert to frequency domain representation
+        HRTF_b = np.fft.fft(brir_b)
```
rkmaddox: Lower case variable names, even when they're acronyms.

expyfun/stimuli/_hrtf.py
```diff
@@ -49,20 +62,56 @@ def _get_hrtf(angle, source, fs):
...
+                         weight_c * np.log10(np.abs(HRTF_c)))
+        HRTF_a_mag = np.power(10, HRTF_a_logmag)
+        HRTF_a_phase = (weight_b * np.unwrap(np.angle(HRTF_b)) +
+                        weight_c * np.unwrap(np.angle(HRTF_c)))
+
+        # combine magnitude and phase components
+        HRTF_a = np.multiply(HRTF_a_mag, np.exp(1j * (HRTF_a_phase)))
+        HRTF_a = _make_sym(HRTF_a)
+        brir_a = np.real(np.fft.ifft(HRTF_a))
+        brir = np.concatenate((brir_a[:, -delay:], brir_a[:, :-delay]), 1)
```
rkmaddox: You can do the shifting and unshifting in place (i.e., replace the variable with the new data rather than making a new one).

expyfun/stimuli/_hrtf.py
Eric89GXL: Yeah I can live with it. The best thing to do would be to have `source=None` -> issue `warnings.warn(..., DeprecationWarning)` for a while and use the old behavior, and eventually change it to `cipic`. But it might be overkill here

rkmaddox: I think it is overkill here. No one who currently uses this library would find that useful.

Eric89GXL:
@rkmaddox @maddycapp27 I'll wait until you converge, feel free to ping me for review once you're ready

expyfun/stimuli/_hrtf.py
maddycapp27: Hmmm, you're right. I think this goes back to the fact that there's no need for interpolation if the angle's hrtf is known.

expyfun/stimuli/_hrtf.py
```diff
@@ -66,49 +66,53 @@ def _get_hrtf(angle, source, fs, interp=False):
...
+    if read_angle in angles:
+        interp = False
     if not interp:
         idx = np.where(angles == read_angle)[0]
         assert len(idx) == 1
         brir = brir[idx[0]].copy()
     elif interp:
         idx = np.where(angles < read_angle)[0]
+        if max(idx) > :
+            raise ValueError('angle "{0}" must be small than "{1}"'
```
rkmaddox: smallER

rkmaddox:
This is now the lone PR. No pressure, though.

expyfun/stimuli/tests/test_stimuli.py
```diff
@@ -31,22 +31,32 @@ def test_textures():
...
+    # invalid angle when interp=False
+    for interp in [True, False]:
+        for source in ['barb', 'cipic']:
+            out = convolve_hrtf(data, 44100, 0, source=source, interp=interp)
+            out_2 = convolve_hrtf(data, 24414, 0, source=source, interp=interp)
+            assert_equal(out.ndim, 2)
+            assert_equal(out.shape[0], 2)
+            assert_true(out.shape[1] > data.size)
+            assert_true(out_2.shape[1] < out.shape[1])
+            if interp==True:
```
drammock: don't test equality of booleans.  `if interp:` is all that is needed here.

expyfun/stimuli/_hrtf.py
```diff
@@ -49,20 +62,60 @@ def _get_hrtf(angle, source, fs):
...
-    idx = np.where(angles == read_angle)[0]
-    assert len(idx) == 1
-    brir = brir[idx[0]].copy()
+    if read_angle in angles:
+        interp = False
+    if not interp:
+        idx = np.where(angles == read_angle)[0]
+        assert len(idx) == 1
+        brir = brir[idx[0]].copy()
+    elif interp:
```
drammock: `interp` is a boolean, no need for `elif interp:` here, just do `else:`

maddycapp27: Thanks!

maddycapp27:
Not sure why appveyor is failing based on the log.

Eric89GXL:
It can be a bit flakey, that's a segfault unrelated to your PR

expyfun/stimuli/_hrtf.py
maddycapp27: Oops, still need to fix that.

expyfun/stimuli/_hrtf.py
```diff
@@ -49,20 +62,56 @@ def _get_hrtf(angle, source, fs):
...
+                             ''.format(angle, max(angles)))
+        # angles are
+        knowns = [angles[idx[-1]], angles[idx[-1] + 1]]
+        # get known brirs
+        brirs = [brir[idx[-1]].copy(), brir[idx[-1] + 1].copy()]
+
+        # find location of maximum for each component of each known brir and
+        # average to find a generally good place to declare time = 0 (all data
+        # from before this point gets shifted to the end)
+        delay = int(np.mean([np.argmax(brirs, 1)]))
```
rkmaddox: I did a bit of messing around with this today and I don't think the delay bit here makes a difference.

maddycapp27: Fair enough--probably cleaner just to delete it.

maddycapp27:
Oh dear, you're right about the negative angles! I'm going to try to see if I can figure it out, and if not...Wednesday.

rkmaddox:
This gist shows the phase-slipping I was talking about. There are frequencies where the phase goes forward by pi when it should go backwards, or vice versa.

https://gist.github.com/rkmaddox/f40b53beb21092a8b50243baec1d70d4

I have a few ideas for how to fix it, but a general solution is non-trivial.

expyfun/stimuli/tests/test_stimuli.py
```diff
@@ -31,22 +31,36 @@ def test_textures():
...
-        out = convolve_hrtf(data, 44100, 0, source=source)[:, 1024:-1024]
-        assert_allclose(np.mean(rms(out)), rms(data), rtol=1e-1)
-        out = convolve_hrtf(data, 44100, -90, source=source)
-        rmss = rms(out)
-        assert_true(rmss[0] > 4 * rmss[1])
+    assert_raises(ValueError, convolve_hrtf, data, 44100, 0.5, interp=False)
+    assert_raises(ValueError, convolve_hrtf, data, 44100, 0, source='other',
+                  interp=False)
+    assert_raises(ValueError, convolve_hrtf, data, 44100, 185, interp=True)
+    assert_raises(ValueError, convolve_hrtf, data, 44100, 0, interp='other')
```
drammock: @maddycapp27 for these `assert_raises` calls, I find it helpful to have either comments explaining what is wrong (cf. the `# invalid angle` comment that used to be here), or alternatively, make it really obvious by the function parameters themselves which one is the invalid one (e.g., by using arguments like `interp='foo'` or `interp='bogus_arg'`).

expyfun/stimuli/_hrtf.py
```diff
@@ -40,29 +39,81 @@ def _get_hrtf(angle, source, fs):
     0-degree BRIRs (mean of that across channels) is equal to 1. This will
     ensure that the RMS of a white signal filtered with this signal is
     unchanged.
+
+    If interp is True, the function takes the two nearest known HRTFs and uses
+    them to calculate an HRTF between them for any given azimuthal angle
+    between 0 and 90 degrees.
+
+    Interpolation relies on averaging log magnitude and phase in the frequency
+    domain as in the referenced paper.
```
Eric89GXL: for proper `sphinx` markup see 

https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

But basically put a `[1]_` here and a `.. [1] R. Martin ...` below in References.

expyfun/stimuli/_hrtf.py
```diff
@@ -40,29 +39,81 @@ def _get_hrtf(angle, source, fs):
...
                          ''.format(angle, list(angles)))
     brir = data['brir']
-    idx = np.where(angles == read_angle)[0]
-    assert len(idx) == 1
-    brir = brir[idx[0]].copy()
+    if read_angle in angles:
+        interp = False
+    if not interp:
+        idx = np.where(angles == read_angle)[0]
+        assert len(idx) == 1
```
Eric89GXL: `assert` statements can be skipped under optimization, check and throw `ValueError`

expyfun/stimuli/_hrtf.py
```diff
@@ -40,29 +39,81 @@ def _get_hrtf(angle, source, fs):
...
+        brir = brir[idx[0]].copy()
```
Eric89GXL: why copy?

expyfun/stimuli/_hrtf.py
```diff
@@ -40,29 +39,81 @@ def _get_hrtf(angle, source, fs):
...
+    else:
+        if source == 'barb':
```
Eric89GXL: a safer (and more readable) check would be `if source != 'cipic'` -- if we ever added more sources this could be a problem

expyfun/stimuli/_hrtf.py
```diff
@@ -40,29 +39,81 @@ def _get_hrtf(angle, source, fs):
...
+            raise ValueError('source must be ''cipic'' when interp=True')
+        idx = np.where(angles < read_angle)[0]
+        if (idx[-1] + 1) > (len(angles) - 1):
```
Eric89GXL: this can fail if `(read_angle > angles).all()`. You might be better off doing a `np.searchsorted` to get the index, then your check can just be `if idx in (0, len(angles)): raise ValueError('... must be between ...')`

expyfun/stimuli/_hrtf.py
```diff
@@ -40,29 +39,81 @@ def _get_hrtf(angle, source, fs):
...
+        hrtf_phase = data['hrtf_phase']
+        pairs = data['pairs']        
+        
+        # isolate appropriate pair of amplitude and phase
+        index = np.where(pairs == knowns)[0][0]
+        hrtf_amp = hrtf_amp[index]
+        hrtf_phase = hrtf_phase[index]
+
+        # weighted averages of log magnitude and unwrapped phase
+        step = knowns[1] - knowns[0]
```
Eric89GXL: `angles` should really be `np.float64` but in case we ever do something like make it an `int`, it's safest to cast this to `float(knowns[1] - knowns[0])`.

expyfun/stimuli/_hrtf.py
```diff
@@ -40,29 +39,81 @@ def _get_hrtf(angle, source, fs):
...
+        weights = (step - np.abs(read_angle - knowns)) / step
+        hrtf_amp = (hrtf_amp[0] ** weights[0] *
+                    hrtf_amp[1] ** weights[1])
```
Eric89GXL: use broadcasting (and in-place) whenever possible:
```
hrtf_amp **= weights[:, np.newaxis]
```

expyfun/stimuli/_hrtf.py
```diff
@@ -40,29 +39,81 @@ def _get_hrtf(angle, source, fs):
...
+        hrtf_phase = (weights[0] * hrtf_phase[0] +
+                      weights[1] * hrtf_phase[1])
```
Eric89GXL: same here

expyfun/stimuli/_hrtf.py
```diff
@@ -40,29 +39,81 @@ def _get_hrtf(angle, source, fs):
...
+        
+        # reconstruct hrtf and convert to time domain
+        hrtf = hrtf_amp * np.exp(1j * hrtf_phase)
+        hrtf = _make_sym(hrtf)
```
Eric89GXL: You can avoid this `_make_sym` and `np.real` business and get a speedup at the same time if you use `np.fft.rfft` and `np.fft.irfft` and store the `n_samples` (to disambiguate between equal-length representations). But if it's a pain to re-do the phase and amplitude stuff don't worry about it.

expyfun/stimuli/_hrtf.py
```diff
@@ -78,6 +129,10 @@ def convolve_hrtf(data, fs, angle, source='barb'):
         Source to use for HRTFs. Currently `'barb'` and `'cipic'` are
         supported. The former is default for legacy purpose. The latter is
         recommended for new experiments.
+    interp : bool
+        Parameter to determine whether to restrict use to known HRTF values or 
+        to use an interpolated HRTF for angles not in the source; set to 
+        'False' by default
```
Eric89GXL: don't quote the False

expyfun/stimuli/_write_hrtf_pairs.py
```diff
@@ -0,0 +1,124 @@
...
+        
+        hrtf_phase[-1][0][0, 41:] += -2 * np.pi
+
+#########################################
+# SAVE AS UNWRAPPED PHASE AND AMPLITUDE #
+#########################################
+hrtf_dict = {'hrtf_phase': hrtf_phase, 'hrtf_amp': hrtf_amp, 'fs': fs,
+             'angles': angles, 'pairs': pairs}
+write_hdf5('/Users/Maddy/.expyfun/data/hrtf/pair_cipic_{0}.hdf5'.format(fs),
+           hrtf_dict, overwrite=True)
```
Eric89GXL: This is a script, not part of the module, so it's weird to have it here. You can instead do one of:

1. Turn the whole thing into a function (and probably omit the writing step because it's clear how to take the vars and create the HDF5)
2. Leave it as a script but add a `expyfun/bin/` and `expyfun/bin/stimuli/` and put this under there (and clean it up to remove things specific to your system, etc.).
3. Just put it in a `gist` (least effort; probably what I would do) and just make a link to the `gist` in the source code comments somewhere.

expyfun/stimuli/_hrtf.py
maddycapp27: This was from the original function so I wasn't sure about why it was there either. It can just be deleted?

Eric89GXL: ahh I realized why it's there -- to allow garbage collection of `brir`. might as well leave it

expyfun/stimuli/_write_hrtf_pairs.py
rkmaddox: I vote for option 3. This is something we are likely never going to have to do again, so not much point in cleaning it up.

expyfun/stimuli/_hrtf.py
maddycapp27: Playing around with this and having trouble understanding--how do I handle multiplying them after applying the weights? I tried this:
```
hrt_amp **= weights[:, np.newaxis]
hrtf_amp = hrtf_amp[0] * hrtf_amp[1]
```
but it wasn't numerically equal to existing code and I suspect that adding the second line rather defeats the purpose of broadcasting. Any insights greatly appreciated!

rkmaddox: ``hrtf_amp = np.prod(hrtf_amp ** weights[:, np.newxis], axis=0)`` makes it a one-liner, but I don't see why it would give a different answer than the above.

Eric89GXL: you have `hrt_amp` instead of `hrtf_amp` on the first line...

rkmaddox: But wait, do you need the ``np.newaxis``? If ``hrtf_amp`` and ``weights`` are the same size, it should just be:

```hrtf_amp = np.prod(hrtf_amp ** weights, axis=0)

Comment (maddycapp27) [inline expyfun/stimuli/_hrtf.py]: That makes sense to me as a thing that would work but evidently not???

wrongmultiplication

Comment (rkmaddox) [inline expyfun/stimuli/_hrtf.py]: That looks like some sort of axis order mix-up to me.

Comment (maddycapp27) [inline expyfun/stimuli/_hrtf.py]: Oops, typo happened when retyping over here, not in my code.

Comment (rkmaddox) [inline expyfun/stimuli/_hrtf.py]: Do you need the ``np.newaxis``? Without knowing the shapes of the variables, it seems like you might be able to do ``hrtf_amp = np.prod(hrtf_amp ** weights, axis=0)``.

```

Eric89GXL: `weights` I assumed is shape `(2,)` whereas `hrtf_amp` is shape `(2, N)` in which case you do need the `np.newaxis`

Eric89GXL: Let's work with code we can easily share. This passes for me:
```Python
>>> import numpy as np
>>> w = np.random.rand(2)
>>> x = np.random.rand(2, 3)
>>> np.testing.assert_allclose(x[0] ** w[0] * x[1] ** w[1], np.prod(x ** w[:, np.newaxis], axis=0))
```

rkmaddox: Works for me, too, though not when `x` has negative values. But ``hrtf_amp`` oughtn't, right?

rkmaddox: Looking at the plots some more: definitely an axis order mix-up. It's basically reading in columns rather than rows. Is there some way the byte order could have been screwed up?

maddycapp27: I think I goofed and this error is coming from other changes

Eric89GXL: BTW for debugging these sorts of things, I often go back to something that worked, and then incrementally add things (or even a parallel code path!) and add `np.testing.assert_allclose(..., ...)` in there to ensure equivalence. If it bombs, you can do post-mortem debugging with `import pdb; pdb.pm()` to inspect the variables with e.g. `p x.ravel()` or `p x * y`

maddycapp27:
@Eric89GXL @drammock Comments addressed, tests passed, meets PEP8. Ready for review and merge--thanks much!

expyfun/stimuli/_hrtf.py
```diff
@@ -40,29 +39,82 @@ def _get_hrtf(angle, source, fs):
     0-degree BRIRs (mean of that across channels) is equal to 1. This will
     ensure that the RMS of a white signal filtered with this signal is
     unchanged.
+
+    If interp is True, the function takes the two nearest known HRTFs and uses
+    them to calculate an HRTF between them for any given azimuthal angle
+    between 0 and 90 degrees.
+
+    Interpolation relies on averaging log magnitude and phase in the frequency
+    domain. [1]
```
Eric89GXL: `[1]_`

expyfun/stimuli/_hrtf.py
```diff
@@ -40,29 +39,82 @@ def _get_hrtf(angle, source, fs):
...
+
+    References
+    ----------
+    [1] R.  Martin and K.  McAnally, "Interpolation of Head-Related Transfer
```
Eric89GXL: `.. [1]` (and it's a `sphinx` directive so be sure to also indent the following lines)

expyfun/stimuli/_hrtf.py
```diff
@@ -40,29 +39,82 @@ def _get_hrtf(angle, source, fs):
...
+        weights = (step - np.abs(read_angle - knowns)) / step
+        hrtf_amp = np.prod(hrtf_amp ** weights[:, np.newaxis, np.newaxis],
+                           axis=0)
+        hrtf_phase = np.sum(hrtf_phase * weights[:, np.newaxis, np.newaxis],
+                            axis=0)
+
+        # reconstruct hrtf and convert to time domain
+        hrtf = hrtf_amp * np.exp(1j * hrtf_phase)
+        hrtf = _make_sym(hrtf)
+        brir = np.real(np.fft.ifft(hrtf))
```
Eric89GXL: WDYT about using `np.fft.rfft` instead? Then you don't need `_make_sym`, the `np.real` here, and the `irfft` should be faster

maddycapp27: Given that the fft happens in the data generation for the files that got merged, I was hesitant to change it.

rkmaddox: I think you could actually leave the data as is and just do ``np.irfft`` on the first half of them. You can get the length information you need from the data file because it is the full FFT. I have not used ``rfft`` or ``irfft`` before. Eric does that seem right?

Eric89GXL: Yes, you should just be able to use `irfft` with a proper `n` argument (and maybe slicing if `hrtf` if necessary, but I doubt it)

maddycapp27: I agree that that makes sense, but I'm concerned about the reality of the brirs being half the length of the non-interpolated ones. Perhaps with resampling the two would converge to the same thing, but after all that I'm worried there wouldn't be an increase in efficiency. 
rfft v fft

rkmaddox: The time-domain BRIR that gets output should be the same, excepting precision. You'd take half of the (full) FFT, and calculate the full BRIR using ``irfft``.

rkmaddox: Oh, wait. It seems like that's what you did, from the plot. Something funny going on. I am eating lunch, then I'll head your way to see if we can figure it out.

maddycapp27: Reread the documentation for irfft and figured out my confusion (dumb and somewhat obvious mistake naturally)

rkmaddox: Didn't pull the length from the original FFT?

maddycapp27: yup >.<

expyfun/stimuli/_hrtf.py
drammock: @maddycapp27 I think you still missed this change.  The underscore after `[1]` is necessary for sphinx to create the hyperlink.

maddycapp27: You are absolutely right! My apologies!

drammock:
@Eric89GXL WTF is up with coveralls?  Did they change the way they count lines?  There are precipitous drops in files that aren't touched by this PR.

maddycapp27:
@drammock I noticed that coveralls was acting up as well. Very confusing. But now everything seems to be ok?

drammock:
thanks @maddycapp27.  Note to self: if you don't like what coveralls says, wait 5 minutes.

maddycapp27:
Yay! Thank you!

PR #310, opened 2017-06-07
MRG: Tracker dealer example

maddycapp27:
This is just a simple example for the TrackerDealer. It's pretty similar to the tracker staircase example but please let me know if there are any changes that should be made. Thanks much!


examples/stimuli/tracker_dealer.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+======================
+Tracker Dealer Example
+======================
+
+This file shows how to interleave multiple Tracker objects using
+:class:'expyfun.stimuli.TrackerDealer'
+
+In this case, a modeled human subject generates two curves (one for each trial
+type: 1 & w)
```
rkmaddox: 1 and 2?

examples/stimuli/tracker_dealer.py
```diff
@@ -0,0 +1,80 @@
...
+                   change_criteria, change_rule, x_min, x_max) for i in [0, 1]]
+
+# initialize TrackerDealer object
+tr = TrackerDealer(tr_UD, max_lag, rand)
+
+# Initialize human state
+rng = np.random.RandomState(1)
+
+while not tr.stopped:
+
```
rkmaddox: Don't need a blank line after ``while``, ``if``, etc.

examples/stimuli/tracker_dealer.py
```diff
@@ -0,0 +1,80 @@
...
+    # that time from TrackerDealer 
+    ss, level = tr.get_trial()  
+    ss = sum(ss)
+    
+    tr_UD[ss].respond(rng.rand() < sigmoid(level - true_thresh[ss],
+                                           lower=chance, slope=slope))
+
+# Plot the results
+for i in [0, 1]:
+
```
rkmaddox: Same, no blank line.

maddycapp27:
Sure thing, here are the plots:
plot1
plot2

examples/stimuli/tracker_dealer.py
maddycapp27: oops...

rkmaddox:
Cool, looks like it's working. They're probably better as subplots so only one plot is made, and that can serve as the thumbnail for the documentation.

Eric89GXL:
I have a real-life use case I need to make -- maybe this can be adapted so I can be lazy and copy-paste more easily.

KC wants interleaved adaptive tracking, one from above and one from below. This is a thing, right @rkmaddox? And `TrackerDealer` is meant to do it? Can this example be made to do it if it doesn't already? (If it does, update docs to say so please.)

rkmaddox:
Yes, it could easily do it. You just make two trackers in ``TrackerDealer`` with different start points.

However, while a fine application, it's not what it was made to do. The reason for making ``TrackerDealer`` was so that you could have interleaved identical trackers for different experimental conditions, not different trackers for the same condition.

So I think the example should stay as is, but it should be super easy for you to do what you want. I am more than happy to help with that if it isn't clear from the docs and example how to do it.

Eric89GXL:
FYI @maddycapp27 a better way to get at what @rkmaddox asked about is to look at CircleCI, which builds all examples. You can see the output if you click the link (sign in with GitHub on CircleCI) and then follow the "Artifacts" tab to the HTML output like here:

https://210-11614571-gh.circle-artifacts.com/0/home/ubuntu/expyfun/doc/_build/html/auto_examples/stimuli/tracker_dealer.html

Eric89GXL:
> I am more than happy to help with that if it isn't clear from the docs and example how to do it.

Yes please make a PR to add docs and ideally an example if you or @maddycapp27 have time

rkmaddox:
On the link @Eric89GXL sent, it looks like ":class:’expyfun.stimuli.TrackerDealer’" isn't rendering properly.

rkmaddox:
> Yes please make a PR to add docs and ideally an example if you or @maddycapp27 have time

There are docs for ``TrackerDealer``. Do they need to modified to make it clear that this is possible?

maddycapp27:
The render is my bad--used the wrong `

Eric89GXL:
> Do they need to modified to make it clear that this is possible?

I was concerned that your statement about it not being designed to handle interleaved top-down / bottom-up (or whatever it's called) meant there had to be some special treatment. If no docs are necessary, it would be good from a pedagogical standpoint to have a simple example, at least so that folks like me can get the terminology right :)

rkmaddox:
Yeah, happy to make an example. It will be a very small modification of @maddycapp27's. The docs don't have any wording about how it should be used, so I think your use case is just as legit as what it was designed for. But a separate example is the best way to show that I think.

rkmaddox:
When this PR gets merged, I'll ask @maddycapp27 to make the modifications and submit another PR based on that successful one.

Eric89GXL:
> When this PR gets merged, I'll ask @maddycapp27 to make the modifications and submit another PR based on that successful one.

Awesome, thanks

rkmaddox:
Glad to see that ``TrackerDealer`` is even more useful than it was intended to be :-)

examples/stimuli/tracker_dealer.py
```diff
@@ -0,0 +1,81 @@
...
+while not tr.stopped:
+    # Get information of which trial type is next and what the level is at
+    # that time from TrackerDealer
+    ss, level = tr.get_trial()
+    ss = sum(ss)
+    tr_UD[ss].respond(rng.rand() < sigmoid(level - true_thresh[ss],
+                                           lower=chance, slope=slope))
+
+# Plot the results
+axes = [plt.subplot(211), plt.subplot(212)]
```
Eric89GXL: `axes = plt.subplots(2, 1)[1]`

examples/stimuli/tracker_dealer.py
```diff
@@ -0,0 +1,81 @@
...
+    lines += tr[i].plot_thresh(4, ax=ax)
+
+    lines[0].set_label('Trials')
+    lines[1].set_label('Reversals')
+    lines[2].set_label('Estimated threshold')
+
+    ax.legend(loc='best')
+    ax.set_title('Adaptive track of model human trial type {} (true threshold '
+                 'is {})'.format(i + 1, true_thresh[i]))
+    plt.tight_layout()
```
Eric89GXL: avoid global `plt` commands, prefer `fig.tight_layout()` here

examples/stimuli/tracker_dealer.py
```diff
@@ -0,0 +1,81 @@
...
+
+@author: maddycapp27
+"""
+
+import numpy as np
+from expyfun.stimuli import TrackerUD, TrackerDealer
+from expyfun.analyze import sigmoid
+import matplotlib.pyplot as plt
+
+# define parameters of modeled subject (using sigmoid probability)
```
Eric89GXL: FYI if you want you can make pretty sections in the output like this:

https://github.com/mne-tools/mne-python/blob/master/tutorials/plot_artifacts_correction_filtering.py#L49

that render like this:

https://mne-tools.github.io/dev/auto_tutorials/plot_artifacts_correction_filtering.html#removing-power-line-noise-with-notch-filtering

might be overkill in this example but it's something to consider when writing pedagogical examples

examples/stimuli/tracker_dealer.py
```diff
@@ -0,0 +1,81 @@
...
+stop_rule = 'reversals'
+start_value = 45
+change_criteria = [0, 5]
+change_rule = 'reversals'
+x_min = 0
+x_max = 90
+
+# parameters for the tracker dealer
+max_lag = 2
+rand = None
```
rkmaddox: I think @Eric89GXL usually likes this to be called ``rng``. And I think in this case it is better to seed it, so ``rng = np.random.RandomState(0)``. Mess around with the seed value until you get plots that end up with thresholds within 1 or 2 of their true value. Seeding the RNGs will make the plots the same every time, and having the computed thresholds close to the the true ones in the example will be less confusing for people.

maddycapp27: Ahhh, ok, I was confused by that in the other example. Will fix!

rkmaddox: Oops, probably shouldn't have been ``None`` there, either.

examples/stimuli/tracker_dealer.py
maddycapp27: Thanks, I'll definitely keep that in mind for the future! For this example, I agree that it's not very necessary though.

examples/stimuli/tracker_dealer.py
```diff
@@ -52,7 +52,7 @@ def callback(event_type, value=None, timestamp=None):
                    change_criteria, change_rule, x_min, x_max) for i in [0, 1]]
 
 # initialize TrackerDealer object
-tr = TrackerDealer(tr_UD, max_lag, rand)
+tr = TrackerDealer(tr_UD, max_lag, rng)
 
 # Initialize human state
 rng = np.random.RandomState(1)
```
maddycapp27: Is it stylistically okay that I overwrite rng here? I could see it being confusing to someone just skimming the example.

rkmaddox: Better to do ``rng_human`` and ``rng_dealer`` or something like that.

Eric89GXL:
@rkmaddox feel free to merge when you and the CIs are happy

Eric89GXL:
@drammock feel free to look, too, if you want

examples/stimuli/tracker_dealer.py
```diff
@@ -0,0 +1,81 @@
...
+# define parameters of modeled subject (using sigmoid probability)
+true_thresh = [30, 40]
+slope = 0.1
+chance = 0.5
+
+# define parameters for each tracker (assuming each tracker uses same rules)
+up = 1
+down = 1
+step_size_up = [9, 3]
+step_size_down = [3, 1]
```
drammock: just so I'm clear... the first element of each of these is "initial step size" and the second element kicks in after `change_criteria[1]` reversals have occured?

maddycapp27: Correct--do you think there should be a comment to that effect?

Eric89GXL: +1 for comment

drammock: yeah, a comment would be good (since initially I thought they were different values that were going to be fed into the two different `TrackerUD`s that were getting created).

drammock:
@Eric89GXL the appveyor error seems unrelated...  I went in and tried to re-start that job but either I wasn't empowered to, or just couldn't figure out how.

drammock:
one last comment: when I pull down this branch, the permissions on `tracker_dealer.py` don't match the rest of the examples (it is marked as executable).  @Eric89GXL @rkmaddox do we care?

Eric89GXL:
> do we care?

I'm all for uniformity so if someone gets around to changing it, go ahead. But I wouldn't hold up the PR for it, as I don't think it should have much consequence.

maddycapp27:
Is the permissions thing something I did?

drammock:
cool.  if another example is going to get generated for @Eric89GXL's use case, maybe we can change the permissions on `tracker_dealer.py` at that point.

drammock:
@maddycapp27 yes (though perhaps not intentionally).  The executable bit is in fact the only aspect of file permissions that git keeps track of (it doesn't track user/group permissions).

maddycapp27:
Certainly not intentionally (I'm a bit ignorant about such things). I'm not sure how to fix that.

Eric89GXL:
> (I'm a bit ignorant about such things). I'm not sure how to fix that.

Something like [chmod a-x tracker_dealer.py](http://lmgtfy.com/?q=set+not+executable+linux) :)

And I'm sure there are some around here who will appreciate your (un?)intentional "bit ignorant" pun... (for those that don't, we're talking about setting the *executable bit*)

maddycapp27:
@Eric89GXL Thank you!

Alas not an intentional pun~ I appreciate the explanation :)

Eric89GXL:
Thanks @maddycapp27, looking forward to the next example :)

PR #311, opened 2017-06-08
MRG: 2nd tracker dealer example

maddycapp27:
Here's the second example that @Eric89GXL asked for.


maddycapp27:
@Eric89GXL @drammock I'm still trying to figure out why this PR includes all the commits from the last example, but wanted to put this up anyway so you could see the actual example you asked for.

Eric89GXL:
You were probably on your old branch when you created your new branch, i.e. left out the first command in this two-command sequence:
```
git checkout master
git checkout -b 2nd_tracker_ex
```

examples/stimuli/tracker_dealer_doublesided.py
```diff
@@ -0,0 +1,100 @@
...
+tr = TrackerDealer(tr_UD, max_lag, rng_dealer)
+
+# Initialize human state
+rng_human = np.random.RandomState(1)  # random seed for modeled subject
+
+while not tr.stopped:
+    # Get information of which trial type is next and what the level is at
+    # that time from TrackerDealer
+    ss, level = tr.get_trial()
+    ss = sum(ss)
```
rkmaddox: Ah, this isn't actually how to interact with ``TrackerDealer``. It's easier than that. You can do:

```
_, level = tr.get_trial()
tr.respond(...)
```

The idea is for ``TrackerDealer`` to handle all your interactions with the trackers. When you do ``get_trial``, things change internally so it knows which tracker to pass the response to. And in fact, it is very possible it will get confused if you're doing things to its trackers without its knowledge. In other words, ``TrackerDealer`` makes use of OOP.

Is this how the other example is? If so I was out to lunch on review, and I apologize!

examples/stimuli/tracker_dealer_doublesided.py
```diff
@@ -0,0 +1,100 @@
...
+# The two trackers in this example use all of the same parameters except for
+# the start valueand then are passed into the dealer. After the dealer is
+# created, the type of trial with the start value above or below the true
+# threshold (returned as an index) and trial level for that trial can be
+# acquired.
+
+# initialize two tracker objects--one for each start value
+tr_UD = [TrackerUD(callback, up, down, step_size_up, step_size_down,
+                   stop_criterion, stop_rule, start_value[i],
+                   change_criteria, change_rule, x_min, x_max) for i in [0, 1]]
```
rkmaddox: Rather than loop over ``i``, you can just do something like ``for sv in start_value``.

examples/stimuli/tracker_dealer_doublesided.py
```diff
@@ -0,0 +1,100 @@
...
+# Initializing and Running Trackers
+# ---------------------------------
+# The two trackers in this example use all of the same parameters except for
+# the start valueand then are passed into the dealer. After the dealer is
+# created, the type of trial with the start value above or below the true
+# threshold (returned as an index) and trial level for that trial can be
+# acquired.
+
+# initialize two tracker objects--one for each start value
+tr_UD = [TrackerUD(callback, up, down, step_size_up, step_size_down,
```
rkmaddox: Lower case variable names.

maddycapp27:
@Eric89GXL Very much oops, you are correct. Shall I kill the branch and do a new PR without the extra files?

Eric89GXL:
Yes please. Note that after making the new branch you can do `git cherry-pick #######` to get your relevant commits.

PR #312, opened 2017-06-08
MRG: Tracker example 2

maddycapp27:
2nd tracker example without the extra stuff.


examples/stimuli/tracker_dealer_doublesided.py
```diff
@@ -0,0 +1,99 @@
...
+
+# parameters for the tracker dealer
+max_lag = 2
+rng_dealer = np.random.RandomState(4)  # random seed for selecting trial type
+
+##############################################################################
+# Initializing and Running Trackers
+# ---------------------------------
+# The two trackers in this example use all of the same parameters except for
+# the start valueand then are passed into the dealer. After the dealer is
```
drammock: missing space `valueand`

examples/stimuli/tracker_dealer_doublesided.py
```diff
@@ -0,0 +1,99 @@
...
+# initialize TrackerDealer object
+tr = TrackerDealer(tr_ud, max_lag, rng_dealer)
+
+# Initialize human state
+rng_human = np.random.RandomState(1)  # random seed for modeled subject
+
+while not tr.stopped:
+    # Get information of which trial type is next and what the level is at
+    # that time from TrackerDealer
+    __, level = tr.get_trial()
```
drammock: convention is to use a single underscore for unused tuple elements returned by a function.  Here there are two underscores.

maddycapp27:
@drammock Moving the files makes a lot of sense to me, but I'm not sure if it will be confusing since the tracker code is in expyfun.stimuli. @rkmaddox ?

Eric89GXL:
I think `experiments` is probably better. The trackers are just a way of choosing stimuli within experiments. The other stuff in `stimuli` is more about how you design stimuli in the first place.

examples/experiments/tracker_dealer_doublesided.py
```diff
@@ -0,0 +1,99 @@
+# -*- coding: utf-8 -*-
+"""
+======================
+Tracker Dealer Example
```
Eric89GXL: name should be different from the other one, perhaps "Adaptive tracking from above and below" (and don't use Title Case)

examples/experiments/tracker_dealer_doublesided.py
```diff
@@ -0,0 +1,99 @@
...
+# initialize TrackerDealer object
+tr = TrackerDealer(tr_ud, max_lag, rng_dealer)
+
+# Initialize human state
+rng_human = np.random.RandomState(1)  # random seed for modeled subject
+
+while not tr.stopped:
+    # Get information of which trial type is next and what the level is at
+    # that time from TrackerDealer
+    _, level = tr.get_trial()
```
Eric89GXL: This is perhaps a question for @rkmaddox but why not use the pattern `for _, level in tr`: (i.e., just make `tr` properly iterable)?

rkmaddox: ``tr`` is iterable. The output of get trial is the index of the tracker and the stimulus level. There is an important point of the design that isn't coming through: the index of the tracker is not returned so that that tracker can be accessed. It's returned so that you know what experimental condition the ``TrackerDealer`` "chose" for you. In other words, it makes the assumption that the trackers it contains are organized in the same way your stimuli/conditions are. That's the reason for ``TrackerDealer``'s ability to have any arbitrary size.

The index it returns should never be used to access the trackers themselves. The reason you can access ``TrackerDealer`` as an array-like is so that you can do things like plot its threshold, etc.

Maybe we should add a flag to ``TrackerUD`` that tells it if it is part of a ``TrackerDealer`` object, and if it is, it doesn't allow you to interact with the actual trackers in inappropriate ways (plotting the data or asking the threshold, yes; logging a response, no)?

Eric89GXL: > The reason you can access TrackerDealer as an array-like is so that you can do things like plot its threshold, etc.

This is a bit different from what I mean.

By "accessing like array-like" I think you mean `__getitem__` i.e. `tr[...]`.

By iterable I mean you should be able to do `for _, level in tr:`, i.e., change what `__iter__` / `iter()` does. I'm saying this shouldn't iterate over `[tr[0], t[1], ...]` (the array-like access), but rather do the iteration that is always used, namely the `while not stopped: _, level = tr.get_trial()` equivalent pattern.

rkmaddox: Ahhh I see. Do you think, since it can be accessed as array-like, that that makes sense? @maddycapp27 and I have already done many commands like ``thresh = [t.threshold() for t in tr]``.

Open to suggestions, since this still isn't in use in a currently running experiment (it is quite close though).

Eric89GXL:
> thresh = [t.threshold() for t in tr]

This could easily be `t.threshold() for t in `tr.trackers` or `for t in tr.iter_trackers()`. Or really the user probably already constructed such a list to pass to `TrackerDealer` in the first place. I think the iteration functionality of `TrackerDealer` should be whatever is going to be used most. I assume this will be the iteration over trials (because it will *always* be used, right?) but maybe not. We could make it `for _, trial in tr.iter_trials():` instead.

Eric89GXL:
... basically the `while not tr.stopped: _, trial = tr.whatever()` should be changed to some form of iteration over `tr` directly or a `tr.iter_something()`. I'm +0.5 on making it `tr` directly for iterating over trials/stimuli. @drammock any sense which you'd prefer?

drammock:
> @drammock any sense which you'd prefer?

Having never used adaptive tracks in my own experiments, I'm finding it hard to keep straight how this is all supposed to work.  I get that the `TrackerUD` keeps track of history internally, decides what level happens on each trial, and decides when to stop.  And I think that the `TrackerDealer` is supposed to "manage multiple instances of `TrackerUD`" but I have never needed to do that so I don't have instincts about what idiom would be most transparent or would cover the most use cases.  If I had to guess naively, I would say that `TrackerUD` should iterate a trial number and a level (which I think it does already), but since I don't really get how the dealer works, I don't know what it should iterate.

rkmaddox:
I am fine with ``t.threshold() for t in tr.trackers``. The other nice thing about that is that it would also move the array-like access from ``tr`` to ``tr.trackers`` (which could then actually *be* an array), discouraging improper use.

Then it would be as you suggest:

```
for condition, level in td:
    x = stimuli[condition]
    # ... present stimulus, get response
    td.respond(correctness)
```

Eric89GXL:
LGTM. @maddycapp27 do you agree?

maddycapp27:
@Eric89GXL I'm also fine with that

Eric89GXL:
@rkmaddox @maddycapp27 I'll let you two work out who wants to make the API changes.

In the meantime, this PR LGTM. @drammock are you happy?

drammock:
sure, on the assumption that when the API gets revised, these examples will also get updated to reflect the revised API, I'm +1 for merging this one as-is.

Eric89GXL:
Thanks @maddycapp27

maddycapp27:
Thanks @Eric89GXL @drammock for all your help!

PR #313, opened 2017-06-09
MRG: Tracker iterable and adding logging for tracker dealer

maddycapp27:
Changes TrackerDealer to iterate through trial information rather than trackers. Edit: Also adding some logging into the TrackerDealer so that it will show up in the tab file (or whatever the callback function dictates) similar to instances of TrackerUD and TrackerBinom.


expyfun/stimuli/_tracker.py
```diff
@@ -705,6 +705,16 @@ class TrackerDealer(object):
...
+    
+    Selects the tracker from which the next trial should be run
+
+    Returns
+    -------
+    subscripts : list-like
+        The position of the selected tracker.
+    x_current : float
+        The level of the selected tracker.
+
```
maddycapp27: This documentation is copied and pasted from the (now deceased) get_trial() function. @Eric89GXL @drammock do you have any thoughts on how to deal with this info?

Eric89GXL: You can put it in the `next` documentation

expyfun/stimuli/_tracker.py
```diff
@@ -705,6 +705,16 @@ class TrackerDealer(object):
     If dealing from TrackerBinom objects (which is probably not a good idea),
     ``stop_early`` must be ``False`` or else they cannot be ensured to keep
     pace.
+
```
Eric89GXL: ... and the `Notes` section above needs to be updated

expyfun/stimuli/_tracker.py
```diff
@@ -738,6 +739,16 @@ def __init__(self, trackers, max_lag=1, rand=None):
         self._tracker_history = np.array([], dtype=int)
         self._response_history = np.array([], dtype=int)
         self._x_history = np.array([], dtype=float)
+        
+        self._dealer_id = id(self)
+        self._callback('dealer_identity', json.dumps(dict(
+            dealer_id=self._dealer_id,
+            tracker_type='dealer')))
```
rkmaddox: ``TrackerDealer`` isn't a tracker, so I wouldn't even include the ``tracker_type`` keyword arg. There are many kinds of trackers, but only one dealer.

expyfun/stimuli/_tracker.py
```diff
@@ -779,46 +805,25 @@ def _pick(self):
...
         """
         if self._trial_complete:
             raise RuntimeError('You must get a trial before you can respond.')
         self._trackers.flat[self._current_tracker].respond(correct)
         self._trial_complete = True
         self._response_history = np.append(self._response_history, correct)
+        if self.stopped:
+            self._callback(
+                'dealer_%i_stop' % self._dealer_id, json.dumps(dict(
+                    tracker_position=[int(s) for s in self._tracker_history],
```
maddycapp27: @Eric89GXL @drammock since what is stored in _tracker_history is actually subscripts, this will obviously bomb for any matrix of trackers with more than one dimension. Is there a better/functional way to format the subscripts?

expyfun/stimuli/_tracker.py
```diff
@@ -738,6 +739,16 @@ def __init__(self, trackers, max_lag=1, rand=None):
         self._tracker_history = np.array([], dtype=int)
         self._response_history = np.array([], dtype=int)
         self._x_history = np.array([], dtype=float)
+        
+        self._dealer_id = id(self)
+        self._callback('dealer_identity', json.dumps(dict(
```
rkmaddox: Check Coveralls, but with callbacks added here, you might need to add tests. Not sure.

expyfun/stimuli/_tracker.py
```diff
@@ -807,6 +818,12 @@ def respond(self, correct):
         self._trackers.flat[self._current_tracker].respond(correct)
         self._trial_complete = True
         self._response_history = np.append(self._response_history, correct)
+        if self.stopped:
+            self._callback(
+                'dealer_%i_stop' % self._dealer_id, json.dumps(dict(
+                    tracker_position=[int(s) for s in self._tracker_history],
+                    responses=[int(s) for s in self._response_history],
+                    x=[int(s) for s in self._x_history])))
```
rkmaddox: Is this the level? If so, there's no guarantee it will be an int, right?

expyfun/stimuli/_tracker.py
rkmaddox: Better to output as is, and call it ``tracker_history``. That should make it the most relatable to what's happening in the tracker. People can later convert the tracker index to a subscript using the ``shape`` tuple from initilization.

expyfun/stimuli/_tracker.py
```diff
@@ -779,46 +805,25 @@ def _pick(self):
...
         if self._trial_complete:
             raise RuntimeError('You must get a trial before you can respond.')
         self._trackers.flat[self._current_tracker].respond(correct)
         self._trial_complete = True
         self._response_history = np.append(self._response_history, correct)
+        if self.stopped:
+            self._callback(
+                'dealer_%i_stop' % self._dealer_id, json.dumps(dict(
+                    tracker_position=[int(s) for s in self._tracker_history],
+                    responses=[int(s) for s in self._response_history],
```
rkmaddox: Call it ``response_history``.

expyfun/stimuli/_tracker.py
rkmaddox: Please call it ``x_history``.

expyfun/stimuli/_tracker.py
maddycapp27: Ohh, I'm reading back in the code and I understand now. Confusion resolved.

expyfun/stimuli/_tracker.py
maddycapp27: In that case, should the` x_history` outputs for each tracker also be called` x_history`?

maddycapp27: Although that would probably require changing other variable names, we should probably leave the TrackerUD and TrackerBinom final logs as is.

rkmaddox: No, the others are ok. This is just to get the JSON logs to have the same names as the parameters.

expyfun/stimuli/tests/test_tracker.py
```diff
@@ -164,12 +163,11 @@ def test_tracker_dealer():
...
     # test array-like indexing
-    dealer_ud[0]
-    dealer_ud[:]
-    dealer_ud[[1, 0, 1]]
-    [d for d in dealer_ud]
+    dealer_ud.trackers[0]
+    dealer_ud.trackers[:]
+    dealer_ud.trackers[[1, 0, 1]]
+    [d for d in dealer_ud.trackers]
     dealer_ud.shape
```
maddycapp27: Does having the shape property return the shape of the trackers still make sense?

expyfun/stimuli/tests/test_tracker.py
```diff
@@ -147,15 +147,14 @@ def test_tracker_dealer():
...
     assert_raises(RuntimeError, dealer_ud.respond, True)
     rand = np.random.RandomState(0)
-    dealer_ud.get_trial()
-    while not dealer_ud.stopped:
-        sub, x_current = dealer_ud.get_trial()
+    
+    for sub, x_current in dealer_ud:
         dealer_ud.respond(rand.rand() < x_current)
         assert(np.abs(dealer_ud[0, 0].n_reversals -
                       dealer_ud[0, 0].n_reversals) <= 1)
```
maddycapp27: @rkmaddox What is this testing?

expyfun/stimuli/tests/test_tracker.py
rkmaddox: Good point. One could get that by doing ``dealer_ud.trackers.shape``. However, since the dealer passes back the subscript for each trial, its shape might be an important enough property that we give people very easy access to it. So I am +0.5 for keeping as is, but I defer to anyone with a strong opinion.

expyfun/stimuli/tests/test_tracker.py
rkmaddox: Heh, I think that should be comparing the number of reversals from two *different* trackers within the dealer, to make sure they never get off by more than 1. But it clearly doesn't do that does it?

expyfun/stimuli/tests/test_tracker.py
Eric89GXL: I'm fine either way

maddycapp27: If it's a question of keep it or toss it, I'd say keep it. As long as there's nothing else we'd like to know the shape of.

expyfun/stimuli/tests/test_tracker.py
maddycapp27: Haha, that makes much more sense! I thought I was missing something!

expyfun/stimuli/tests/test_tracker.py
rkmaddox: That's two +0.5s. I'd say keep it.

expyfun/io/_parse.py
```diff
@@ -6,6 +6,33 @@
...
+        lines = [c for c in csvr]
+
+    # first two lines are headers
+    assert (len(lines[0]) == 1 and lines[0][0][0] == '#')
+    #metadata = ast.literal_eval(lines[0][0][2:])
+    assert lines[1] == ['timestamp', 'event', 'value']
+    lines = lines[2:]
+
+    header = list(set([l[1] for l in lines]))
+    header.sort()
```
drammock: shouldn't this `return` something?

maddycapp27: Oops, that was something else I'm working on, nothing to do with this.

expyfun/stimuli/_tracker.py
```diff
@@ -722,6 +722,10 @@ def __init__(self, callback, trackers, max_lag=1, rand=None):
                 raise ValueError('stop_early for trackers.flat[%d] must be '
                                  'False to deal trials from a TrackerBinom '
                                  'object' % (ti,))
+            if isinstance(t, TrackerBinom) and not t.x_current:
+                raise ValueError('trackers.flat[%d] must have x_current value'
```
drammock: need space at end of this string (or at beginning of string on next line)

expyfun/stimuli/_tracker.py
```diff
@@ -820,9 +820,9 @@ def respond(self, correct):
         if self.stopped:
             self._callback(
                 'dealer_%i_stop' % self._dealer_id, json.dumps(dict(
-                    tracker_position=[int(s) for s in self._tracker_history],
-                    responses=[int(s) for s in self._response_history],
-                    x=[int(s) for s in self._x_history])))
+                    tracker_history=[int(s) for s in self._tracker_history],
+                    response_history=[int(s) for s in self._response_history],
+                    x_history=[float(s) for s in self._x_history])))
```
maddycapp27: Using `float(s)` now makes it unhappy for `Tracker_Binom` where `x_current=None`

rkmaddox: Hmmmmmm. Perhaps we should set ``x_current`` to ``np.nan`` rather than ``None`` for ``Tracker_Binom``. It doesn't matter too much, since the point is that ``Tracker_Binom`` doesn't actually use that paramter.

Eric89GXL:
Now that I'm starting to try to use this, a couple of small API questions:

1. Why `change_criteria` and `change_rule` instead of `change_reversals=None, change_trials=None` (you can set one or both)?

2. For step_size_up / step_size_down lists, shouldn't `change_criteria` be 1 less than the length of these, since you only need to make N-1 change-size transitions for N step sizes?

Eric89GXL:
I think I got my answer to (2):
```
  File "/home/larsoner/Documents/python/larsoner/adaptivetempcoh/adaptivetempcoh.py", line 65, in <module>
    change_rule, x_min, x_max) for sv in start_value]
  File "/home/larsoner/Documents/python/expyfun/expyfun/stimuli/_tracker.py", line 134, in __init__
    raise ValueError('First element of change_criteria must be 0.')
ValueError: First element of change_criteria must be 0.
```
I'd rather it just be N-1 in length...

Eric89GXL:
And for (1), another more flexible API would be e.g., `stop=dict(reversals=30, trials=20)`, `change=dict(reversals=[5, 2])`

Eric89GXL:
Sorry I didn't catch these on the first go-around. We can stick with the current API if it makes your lives difficult to change them.

examples/experiments/tracker_dealer_doublesided.py
```diff
@@ -69,25 +69,24 @@ def callback(event_type, value=None, timestamp=None):
...
 
 ##############################################################################
 # Plotting the Results
 # ---------------------------
 axes = plt.subplots(2, 1)[1]
 for i in [0, 1]:
-    fig, ax, lines = tr[i].plot(ax=axes[i])
-    lines += tr[i].plot_thresh(4, ax=ax)
+    fig, ax, lines = td.trackers.ravel()[i].plot(ax=axes[i])
+    lines += td.trackers.ravel()[i].plot_thresh(4, ax=ax)
```
Eric89GXL: BTW all of the `set_label` calls below should probably live in `tracker.plot()`. When doing `ax.plot(...)` you can do `ax.plot(..., label=...)`.

rkmaddox:
It's not a big deal to make changes right now. The first element must be 0 thing was a choice I wrestled with a bunch. So if you prefer it the other way I'm OK with that.

In principle I like ``stop=dict(reversals=30, trials=20)``. That's actually pretty nice. Almost always people use one or the other, but I see no reason not to make it more flexible. That ``change`` dict though is weird. I cannot imagine a case where you would want to change based on trials *or* reversals. It would be super messy. But I guess we could change it for consistency.

The thing that gets complicated is that ``TrackerDealer`` needs to keep pace on either trials or reversals. It can't do both. So if you have two stop rules, things can get weird. It also currently uses ``stop_rule`` for pacing because that's what makes the tracks stop around the same time without having a bunch of trials from one condition left to finish up. However, the pacing code as written should be able to handle the case where one tracker needs to be finished up.

I would say to minimize the load on @maddycapp27 and keep target movement to a minimum, poke around a bit more and comment here for things you think should be changed, and then we can decide and address them all in one go. Sound OK?

maddycapp27:
Happy to make whatever changes! (Happy in a "it will happen sometime next week" way).

Eric89GXL:
Yeah that's fine. One reason `dict` is nice is that if there is ever another reason we want to add, we can do it without parameter explosion. But if that's unlikely (probably is), then the `stop_reversals=None, stop_trials=None` (where you mest define exactly one) might be the best way to go.

Also I'm not sure if these counts are working the way I'd expect. I wanted a step size of 4 for 3 trials, 2 for 7 trials, and 1 for the remainder:
```
step_size_up, step_size_down = [4, 2, 1], [4, 2, 1]
change_criteria = [0, 3, 7]
stop_criterion = 30
stop_rule = 'reversals'
x_min, x_max = 1, 20
start_value = [x_min, x_max]
```
But this was the track (fixed response True @ `level > 10`):

screenshot from 2017-06-09 16-04-59

rkmaddox:
The elements of ``change_criteria`` are indices, not counts. (That's why the first one is currently required to be 0). It looks like in your tracks, you are getting changes at 3 and 7 reversals.

rkmaddox:
So, try ``change_criteria = [0, 3, 10]`` and ``change_rule = reversals`` and see if that does what you want.

Eric89GXL:
Counts seems more intuitive to me than indices. To do indices I'd actually figure out the counts I wanted and `cumsum`.

Eric89GXL:
Ahh yes I see it's properly documented already and I just didn't do my homework. My bad!

But do you find counts to be at least as intuitive? Either way you'll have to do a `diff` or `cumsum` operation to get between the two representations. But with the `counts` rather than `indices` we at least get rid of that pesky 0 :)

rkmaddox:
Nah, actually to me it seems more intuitive to say it changed at X and Y reversals. But I don't honestly care that much (it was another thing that I debated over when initially coding it).

maddycapp27:
I think I'm pro-indices--makes it easier to think about how "soon" to change step size based on the stop condition.

Eric89GXL:
Fair enough, we can keep that. How about:

1. change the name to `change_indices`, because they are indices into the reversals (or trials)
2. remove the zero, enforce list length = step size length -1

Then I don't think I would have needed to look at the docs to get it right :)

rkmaddox:
Sure, if @maddycapp27 is OK with it.

examples/experiments/tracker_dealer.py
```diff
@@ -40,7 +40,7 @@
 stop_criterion = 30
 stop_rule = 'reversals'
 start_value = 45
-change_criteria = [0, 5]
+change_criteria = [5]
```
drammock: might as well change these variable names to `change_indices` too.

examples/experiments/tracker_dealer.py
```diff
@@ -66,7 +66,7 @@ def callback(event_type, value=None, timestamp=None):
 # initialize two tracker objects--one for each trial type
 tr_ud = [TrackerUD(callback, up, down, step_size_up, step_size_down,
                    stop_criterion, stop_rule, start_value,
-                   change_criteria, change_rule, x_min, x_max) for i in [0, 1]]
+                   change_criteria, change_rule, x_min, x_max) for i in [0, 1, 2]]
```
drammock: here too, `change_criteria` -> `change_indices` (or `change_index`)

examples/experiments/tracker_dealer_doublesided.py
```diff
@@ -39,7 +39,7 @@
 stop_criterion = 30
 stop_rule = 'reversals'
 start_value = [15, 45]
-change_criteria = [0, 5]
+change_criteria = [5]
```
drammock: and here.

expyfun/stimuli/_tracker.py
```diff
@@ -103,21 +103,20 @@ class TrackerUD(object):
...
 
         ..., step_size_up=[1., 0.2], step_size_down=[1., 0.2], \
-        change_criteria=[0, 2], change_rule='reversals', ...
+        change_indices=[2], change_rule='reversals', ...
 
     would change the step sizes from 1 to 0.2 after two reversals.
 
     If static step sizes are used, both ``step_size_up``
-    and ``step_sizedownp`` must be scalars and ``change_criteria`` must be
+    and ``step_sizedownp`` must be scalars and ``change_indices`` must be
```
drammock: `step_sizedownp` -> `step_size_down`

expyfun/stimuli/_tracker.py
```diff
@@ -779,46 +814,25 @@ def _pick(self):
...
         if self._trial_complete:
             raise RuntimeError('You must get a trial before you can respond.')
         self._trackers.flat[self._current_tracker].respond(correct)
         self._trial_complete = True
         self._response_history = np.append(self._response_history, correct)
+        if self.stopped:
+            self._callback(
+                'dealer_%i_stop' % self._dealer_id, json.dumps(dict(
+                    tracker_history=[int(s) for s in self._tracker_history],
+                    response_history=[float(s) for s in self._response_history],
```
maddycapp27: Ended up changing the `response_history` in the log because trying to convert them to `int` was an issue with the `nan`s

Eric89GXL:
Flake8

https://travis-ci.org/LABSN/expyfun/jobs/242121725

Failing example(s)

https://circleci.com/gh/maddycapp27/expyfun/49?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

maddycapp27:
I think the move from `change_criteria` to `change_indices` is finally debugged (although the tests will have the final say on that once they find). @Eric89GXL @drammock Looking at your comments the other potential changes were:

1. Offering both a reversal _and_ a trial stopping criteria: Is this a "whichever happens first" or "both must happen" kind of deal? Not adverse to having this functionality, but not sure how high priority the change is based on previous discussion.
2. Putting labels in the plot function: How do you feel about the plot function also calling the `plot_thresh` function by default (perhaps include an argument of `threshold=True`)?

Was/is there anything else?

rkmaddox:
1. My initial reaction was *definitely either*, then I hesitated, then I settled at "either, right?" So: either, right?

2. I feel fine about it if there's a flag.

drammock:
1. regarding the reversals: would it work to have a `min_trials`, `max_trials`, and `min_reversals` parameters, where defaults are `min_trials=0, max_trials=np.inf`?  That way you just check against all 3 params each time, the default case responds only to `min_reversals`, but it's easy to put in trial ceiling/floor as desired.

2. agree with @rkmaddox: fine if there's a flag

maddycapp27:
1. Hmm, I like that. In the case where you hit `max_trials` before `min_reversals` I would say that the tracker should stop (i.e. `max_trials` carries more importance than `min_reversals`). Agree/disagree?
2. Dumb question: what do you mean by flag?

drammock:
"flag" = function parameter (usually a boolean one).  Meaning you can turn off the behavior if you don't want it, i.e., by passing `plot_thresholds=False`

drammock:
by the way, I notice all the commits on this thread say they're coming from @rkmaddox but based on the conversation it sounds like @maddycapp27 is actually doing the coding.  Seems like maybe you're working on a shared computer and didn't update the `git config`?  (Might be an argument in favor of having separate user accounts on the machine...)

maddycapp27:
Yes you are correct, I should have made it clear that I'm doing the changes. We'll figure out an arrangement to fix the ambiguity soonish.

maddycapp27:
I ended up making parameters for `stop_reversals` and `stop_trials` instead of doing min/max_trials and min_reversals. The latter was making modification to the `TrackerDealer` rather arduous (read: frustrating/time consuming).

Eric89GXL:
Don't forget to ping for review when changing to MRG, as there is not title-change GitHub notification

expyfun/stimuli/_tracker.py
```diff
@@ -369,13 +385,16 @@ def reversal_inds(self):
...
+    def plot(self, ax=None, threshold=True, n_skip=2):
         """Plot the adaptive track.
 
         Parameters
         ----------
         ax : AxesSubplot | None
             The axes to make the plot on. If ``None`` defaults to current axes.
+        threshold : bool
+            Whether to plot the estimated threshold on the axes. Default is
+            True.
```
Eric89GXL: n_skip is not documented

expyfun/stimuli/_tracker.py
```diff
@@ -491,7 +516,7 @@ class TrackerBinom(object):
     stop_early : boolean
         Whether to stop the adaptive track as soon as ``alpha`` is reached and
         at least ``min_trials`` have been presented.
-    x_current : float | None
+    x_current : float | np.nan
         The level that you want to run the test at. This has no bearing on how
         the track runs, and it will never change, but it is required to be
         here for ``TrackerDealer``.
```
Eric89GXL: `nan` is already a `float`, no need to distinguish

I'm a bit confused by this. Why does it need to be a parameter if it has no bearing on the values / outcomes? Is this from an earlier version back when `TrackerDealer` would do the initialization of its trackers (which is no longer supported I think)?

expyfun/stimuli/_tracker.py
```diff
@@ -259,18 +268,25 @@ def respond(self, correct):
...
 
     def _step_index(self):
         if self._change_rule.lower() == 'reversals':
             self._n_change = self._n_reversals
-        elif self._stop_rule.lower() == 'trials':
+        elif self._change_rule.lower() == 'trials':
             self._n_change = self._n_trials
-        return np.where(self._n_change >= self._change_criteria)[0][-1]
+        step_index = np.where(self._n_change >= self._change_indices)[0]
+        if len(step_index) == 0 or self._change_indices == [0]:
```
Eric89GXL: I get an error on this line:
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
This is because `self._change_indices == np.array([4, 8])`. Please fix and add a test that would have exposed this error.

Eric89GXL: (You probably want `np.array_equal(change_indices, [0])` or `self._change_indices[0] == 0` or `len(self._change_indices) == 0` or so)

Eric89GXL:
Okay I actually did a track while half paying attention to the (poorly-constructed) stimuli:

screenshot from 2017-06-14 12-59-25

You can see for a while in a couple of them I was actually at ceiling (level 1) performance. I'm thinking that these should perhaps effectively count as reversals...? In the limit where a subject really is at ceiling based on the stimuli, then it seems like the tracker should be able to capture that, and it would if these counted as reversals.

Eric89GXL:
Also, perhaps at least the x-axes of all trackers should be linked (so they're on the same scale), and then make linking the y-axes optional (since for from-above and from-below it can be nice for camparing)?

Eric89GXL:
> then make linking the y-axes optional (since for from-above and from-below it can be nice for camparing)?

Actually perhaps the options should be `ylim='auto'` (current) or `'extrema'` (new default that uses min/max?) or `'linked'` (which is 'auto' but with mpl linking).

This should probably be a separate PR since it will involve some thinking, and I don't want to hold up this PR for it.

expyfun/stimuli/_tracker.py
```diff
@@ -738,19 +770,43 @@ def __init__(self, trackers, max_lag=1, rand=None):
...
+        if not self._trial_complete:
+            # Chose a new tracker before responding, so record non-response
+            self._response_history = np.append(self._response_history,
+                                               np.nan)
+        self._trial_complete = False
+        self._current_tracker = self._pick()
+        self._tracker_history = np.append(self._tracker_history,
+                                          self._current_tracker)
+        ss = np.unravel_index(self._current_tracker, self.shape)
+        level = self._trackers.flat[self._current_tracker].x_current
```
maddycapp27: @Eric89GXL This is why `TrackerBinom` needs to have an `x_current` property even if you're using `np.nan`--but are you saying that we should remove the flag for it in `TrackerBinom`?? I'd be hesitant to do so as including it is a nice way to record what level you did the testing at.

rkmaddox: +1

Eric89GXL: Okay if that's the case update the docs to reflect this -- currently they say it's useful in a way that makes it sound like TrackerDealer's API needs it (which the user does not care about), but it's (also) useful so that the level gets documented properly (which the user will care about).

rkmaddox:
@Eric89GXL with the plotting options, where would these go? The only plotting function is in the tracker, not the dealer, so having ``linked`` as an option wouldn't make any sense. But, I agree that it should be saved for another PR. There are plenty of enhancements that can be made (I want to add a ML psychometric function fit, for example) that don't affect how it is used in the experiment. We need to get that down so that @maddycapp27 can move forward, and then we can add ENH PR's as we use it some more.

rkmaddox:
Also, per an ongoing email conversation with Stecker, if you're bouncing off your upper difficulty limit a lot then your steps are too big.

Eric89GXL:
> Also, per an ongoing email conversation with Stecker, if you're bouncing off your upper difficulty limit a lot then your steps are too big.

Well I wasn't really bouncing so much as staying there. That's why I think they should count as "reversals". They would:

1. Help you get to the finer resolution more quickly if using multiple sizes, and (more critically)
2. Prevent infinite trial numbers if someone really can do the hardest difficulty consistently.

Eric89GXL:
If we don't want to count staying at ceiling or floor from trial-to-trial as reversals, then we probably need to make the `stop_trials` argument mandatory (non-None), lest some poor subjects be left doing the task forever.

rkmaddox:
Hmm. If there are subjects that can consistently perform the task at the hardest difficulty level, then I think an adaptive track probably isn't the right way to find the threshold. That would mean that the threshold was somewhere below the hardest condition, and regardless of what you count as reversals, the threshold you calculate will be meaningless.

I think for this reason I am against counting them as reversals. It would basically speed up and disguise getting a meaningless answer. I think we can assume a reasonable amount of experimenter care that should prevent those edge cases from happening.

maddycapp27:
I don't love the idea of forcing `stop_trials` to be `int`. For me, I question if this would prematurely end a run more often than it aborted an endless run. But I also can't think of a better solution.

Putting a warning in seems reasonable. I would imagine that you wouldn't want to warn at a case like trial 7 in the second Run of your example, but would specifically want to warn about cases where there were multiple consecutive ceiling/floor responses.

rkmaddox:
Seems to me like the solution is setting ``stop_trials`` to be a finite ``int`` if you're concerned that a run might take too long. Especially since ``np.inf`` can't be cast to ``int``, in cases where you don't want an upper limit on trials, you have to do some arbitrarily large ``int``, or allow a flag like -1 or something, both of which seem messy. We could add to the docs the possibility of a track taking too long, or a poorly constructed one never converging.

But a warning for threshold seems fine. Eventually we will want to add plan to add a different threshold function that uses the the ML PF fit, rather than averaging reversals. The latter is a quick and dirty estimate.

rkmaddox:
I am still pondering the reversals thing.

Eric89GXL:
> I don't love the idea of forcing stop_trials to be int. For me, I question if this would prematurely end a run more often than it aborted an endless run.

Well then we can disallow `None`, and force people to do `np.inf` if they are confident enough (but recommend in the docs that they don't do this unless they're certain no participant can achieve ceiling perf).

> Especially since np.inf can't be cast to int, in cases where you don't want an upper limit on trials, you have to do some arbitrarily large int, or allow a flag like -1 or something, both of which seem messy. 

We can cast to `float` if want `np.inf` to be a possibility for those experimenters who are sure subjects can never ceiling out.

> I am still pondering the reversals thing.

The track I posted is actually an interesting case. My thresh is probably more like 3 than 1 in my task. I actually *know* that 1 is chance level by construction -- it's a 2I-2AFC task where intervals 1 and 2 only meaningfully differ if level > 1.

I'm also not sure the "right" way to handle these things, because they can happen by chance in 2AFC tasks where there is some upper limit on the stimulus difficulty level.

Maybe the best thing to do for now (to get this PR in!) is to force `stop_trials` to be `int | float`, recommend people pass an `int` to prevent infinite length in the case of ceiling performance, but allow them to use `np.inf` if they want. That way we can keep thinking about how to deal with repeated ceiling performance.

maddycapp27:
> Well then we can disallow `None`, and force people to do `np.inf`

`stop_trials` is recast as `np.inf` if it's `None` in the function call--is there significance to having the user input `np.inf` explicitly?

Eric89GXL:
> stop_trials is recast as np.inf if it's None in the function call--is there significance to having the user input np.inf explicitly?

It's not a big difference, but it is a little bit more explicit IMO in terms of forcing people to realize they could theoretically have an infinite number of trials. But if you prefer `None`, I can live with it as long as we're clear it's the same as `np.inf`, i.e. it could be infinitely long in the docs.

rkmaddox:
@maddycapp27 Functionally, no, but it forces the experimenter to ponder the infinite.

rkmaddox:
For the reversals thing, whether or not it is a good idea will depend on how you calculate your threshold. If you are doing mean of reversals (the standard, but not great way), then your threshold won't be affected if you immediately reverse at ceiling, so it's fine. If you average all levels (less standard, but slightly better) this will really screw up your estimate. However, it was a bad track anyway. You also probably won't be able to do a great ML fit, but I don't think calling it a reversal will help.

So I think it is OK to call it a reversal when you hit the floor/ceiling? Because that is basically the level you will stay at until you get a real reversal anyway. Does that seem right, both?

Eric89GXL:
Example broken

https://circleci.com/gh/maddycapp27/expyfun/62?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Thanks for sticking with this @maddycapp27 I think we're almost there

maddycapp27:
> I think we're almost there

Fingers crossed!

Eric89GXL:
> So I think it is OK to call it a reversal when you hit the floor/ceiling? Because that is basically the level you will stay at until you get a real reversal anyway. Does that seem right, both?

Yeah I'm fine with that. We could add a param that is `repeat_limit='reversal'` (new default) or `'ignore'` (current behavior), unless you thinks YAGNI and we can just effectively use `at_ceiling='reversal'` without exposing the param.

rkmaddox:
Let's do ``repeat_limit='reversal'`` without exposing the parameter for now, if that's OK with @maddycapp27.

maddycapp27:
> Let's do repeat_limit='reversal' without exposing the parameter for now, if that's OK

In this PR?

rkmaddox:
I think this PR, at least here on out, should be focused on nailing down the behavior of the tracker during an experiment. Any other plotting or analysis things that come up can be done later. So I think ``repeat_limit`` should be in this PR, since it will in fact affect experiment flow.

rkmaddox:
(Sorry. This has kind of become the PR from hell)

Eric89GXL:
> (Sorry. This has kind of become the PR from hell)

On the contrary. Consider yourself lucky by considering the two interrelated [cautionary](https://github.com/mne-tools/mne-python/pull/3310) [tales](https://github.com/mne-tools/mne-python/pull/3728) from MNE which together contain > 1000 comments. @rkmaddox was also involved in those, just sayin'.

I did learn an [important link](https://www.youtube.com/watch?v=6b9ci_z4v7M&feature=youtu.be&t=2m46s) from those, though. (If you maintain a GitHub repository, there is a big green "Merge" button to click to accept changes.)

rkmaddox:
Maybe that means any PR I get involved in goes to hell?

Eric89GXL:
The most-commented PR on this repo is #301...

rkmaddox:
I just deleted that branch for closure.

expyfun/stimuli/_tracker.py
```diff
@@ -176,10 +190,10 @@ def __init__(self, callback, up, down, step_size_up, step_size_down,
             down=self._down,
             step_size_up=[int(s) for s in self._step_size_up],
```
rkmaddox: float

expyfun/stimuli/_tracker.py
```diff
@@ -176,10 +190,10 @@ def __init__(self, callback, up, down, step_size_up, step_size_down,
...
             step_size_down=[int(s) for s in self._step_size_down],
```
rkmaddox: float

expyfun/stimuli/_tracker.py
```diff
@@ -259,18 +277,25 @@ def respond(self, correct):
                     x=[int(s) for s in self._x])))
```
rkmaddox: float

drammock:
just FYI, I'm not likely to be able to review this...  super busy the rest of this week and next week, and then off to Kavli Institute for 2 weeks.  Forge ahead without me!

rkmaddox:
@drammock That's awesome, enjoy! Too bad you're not heading up UR way, we're only a few hours north.

maddycapp27:
@Eric89GXL I'm good. Thanks :)

expyfun/stimuli/_tracker.py
```diff
@@ -223,6 +237,10 @@ def respond(self, correct):
                     self._n_reversals += 1
                 if self._direction <= 0:
                     self._direction = 1
+        if self._repeat_limit == 'reversal':
```
rkmaddox: This should also do a step, right? So if you're at the floor, count it as a reversal and move up. Otherwise there's reversals with no actual change in direction.

expyfun/stimuli/_tracker.py
```diff
@@ -60,28 +60,33 @@ class TrackerUD(object):
...
-        the same. If list of float then it will change when ``change_criteria``
+        the same. If list of float then it will change when ``change_indices``
         are encountered. See note below for more specific information on
         dynamic tracker parameters specified with a list.
-    stop_criterion : int
-        The number of reversals or trials that will stop the tracker.
-    stop_rule : str
-        How to determine when the tracker stops. Can be 'reversals' or
-        'trials'.
+    stop_reversals : int | None
```
rkmaddox: Do the ``np.inf`` stuff to make it mimic ``stop_trials``

expyfun/stimuli/_tracker.py
rkmaddox: Does that seem right to you guys?

Eric89GXL: I think our choices here are to:

1. Stay at the same level. This means it counts as a reversal despite no change in direction.
2. Go away from the limit. This means there is a change in direction, but it is in the opposite direction of what is suggested by the behavioral response.

I think I actually prefer (1). It's clearer, then, that changes in direction are only ever caused by a behavioral response that indicates moving in that direction.

I'm not sure how this border behavior works with e.g. 2-up 1-down, though -- I've only been thinking in the context of 1-up 1-down.

maddycapp27: I'm with @Eric89GXL -- if hitting a min is going to bias the "true threshold" up I feel like changing the direction will make it worse.

rkmaddox: Yeah I guess I am OK with that. Really you should not be calculating the threshold from reversals or trials if you are more than brushing up against the min every now and then.

rkmaddox:
I think I am +1 for merge. It seems to have converged on experimental flow, at least for a properly set up track.

Eric89GXL:
Thanks for working through this @maddycapp27 !

Eric89GXL:
So `_repeat_limit = 'reversal'` is set but never used. Can you make a PR to fix this behavior?

Eric89GXL:
(i.e., this PR didn't do what we had talked about it doing when at an extreme, namely treat it as a reversal)

maddycapp27:
I can look at it tomorrow if that's ok?

Eric89GXL:
Sure, that would be good

PR #314, opened 2017-06-09
MRG: manage texture verbosity

Eric89GXL:
I am generating a ton of these stimuli, and don't need to hear over and over that they are 1.45 ERB apart.


drammock:
@Eric89GXL same old appveyor random failure (`test EC with TDT` on the 32-bit build).  Merging anyway.

PR #315, opened 2017-06-19
MRG: Small change to TrackerBinom docs

maddycapp27:
Upon closer inspection, it looks like TrackerBinom isn't actually doing what we said it was doing. This PR is just to make the docs make sense.


rkmaddox:
Thanks for the fix.

PR #316, opened 2017-06-19
MRG: add read_tab_raw and reconstruct_tracker

maddycapp27:
Because the current `read_tab` function does not allow for much customization of how tab files are read, this PR is to introduce two new functions:

`read_tab_raw` : The most basic tab reading function. The goal is to just take a tab file and spit out a list of tuples (where each tuple is one line from the original file--excluding the first two lines (which are header))

`reconstruct_tracker` : This will allow tracker objects to be reconstructed from experiment logs such that analysis functions of the tracker (such as `plot`) can be used.


maddycapp27:
This is very much a messy first attempt, but I wanted to go ahead and open up a PR so that discussion on what each of these functions _should_ do can happen.

expyfun/io/_parse.py
```diff
@@ -73,3 +100,37 @@ def read_tab(fname, group_start='trial_id', group_end='trial_ok'):
...
+                tr.append(TrackerUD(**t_dict))
+            else:
+                tr.append(TrackerBinom(**t_dict))
+            stop_str = 'tracker_' + str(tracker_id) + '_stopped'
+            tracker_stop_idx = np.where(raw[1] == stop_str)
+            responses = ast.literal_eval(raw[tracker_stop_idx][2]['responses'])
+            # feed in responses from tracker_ID_stop
+            [tr.respond(r) for r in responses]
+            
+        return tr
```
rkmaddox: Need a new line at end of file (PEP8 thing). I'll come over for a demo in a few minutes.

expyfun/io/_parse.py
```diff
@@ -4,6 +4,39 @@
 
 import numpy as np
 import csv
+import ast
+import json
+from expyfun.stimuli import TrackerUD, TrackerBinom, TrackerDealer
```
maddycapp27: Somehow I'm not having any luck importing the tracker objects. I tried 
`from expyfun.stimuli import TrackerUD, TrackerBinom, TrackerDealer` and `from ..stimuli import TrackerUD, TrackerBinom, TrackerDealer` but am getting errors for both cases (_ValueError: Attempted relative import in non-package_ and _ImportError: cannot import name TrackerUD_ respectively). Any thoughts?

Eric89GXL: You might have a circular import problem. Nest the import to `Tracker*`, and to be clean make it a relative import, and you should be good.

expyfun/io/_parse.py
```diff
@@ -73,3 +99,53 @@ def read_tab(fname, group_start='trial_id', group_end='trial_ok'):
             d[key] = [(these_vals[ii], these_times[ii]) for ii in idx]
         data.append(d)
     return data
+    
+    
+    def reconstruct_tracker(fname):
+        # read in raw data
+        raw = read_tab_raw(fname)
```
Eric89GXL: i.e., put `from ..stimuli import TrackerUD, ...` as the first line in this function.

And did you really mean for this function to be nested in `read_tab`?

expyfun/io/_parse.py
rkmaddox: Whoops, forgot to come over. I'll head over now.

expyfun/io/_parse.py
maddycapp27: Everything is better now :)

expyfun/io/_parse.py
maddycapp27: My intent is for read_tab_raw to be the basic function that can either lead into read_tab (for people who want to split up trials) or to reconstructing tracker objects (where splitting up trials makes less sense.)

Eric89GXL: No no no -- I'm asking about why `def reconstruct_tracker` is nested in `read_tab` after the `return` statement of `read_tab`, where it can never be reached... i.e. I think you have an indentation problem

maddycapp27: That is definitely an oops.

expyfun/io/tests/test_parse.py
```diff
@@ -45,3 +46,8 @@ def test_parse():
     data = read_tab(ec.data_fname, group_end=None)
     assert_equal(len(data[0]['misc']), 2)  # includes between-trials stuff
     assert_equal(len(data[1]['misc']), 2)
+
+def test_reconstruct_tracker():
+    """Test reconstruction of TrackerUD and TrackerBinom objects from .tab
+    log files."""
+    with ExperimentController(*std_args, stim_fs=44100, **std_kwargs) as ec:
```
maddycapp27: This is only half a thought, I will definitely get back to it.

expyfun/io/tests/test_parse.py
```diff
@@ -48,13 +48,37 @@ def test_parse():
     assert_equal(len(data[0]['misc']), 2)  # includes between-trials stuff
     assert_equal(len(data[1]['misc']), 2)
 
-def test_reconstruct_tracker():
-    """Test reconstruction of TrackerUD and TrackerBinom objects from .tab
-    log files."""
-    from expyfun.io import reconstruct_tracker
+def test_reconstruct():
+    """Test Tracker objects reconstruction"""
+    from expyfun.io import reconstruct_tracker, reconstruct_dealer
```
maddycapp27: This is yielding me another fun circular import problem. Still pondering how to fix it.

maddycapp27: Or it's just not recognizing that the new functions in the parse file exist.

expyfun/io/tests/test_parse.py
```diff
@@ -45,3 +46,49 @@ def test_parse():
     data = read_tab(ec.data_fname, group_end=None)
     assert_equal(len(data[0]['misc']), 2)  # includes between-trials stuff
     assert_equal(len(data[1]['misc']), 2)
+
+def test_reconstruct():
+    """Test Tracker objects reconstruction"""
+    from expyfun.stimuli import TrackerUD, TrackerBinom, TrackerDealer
+    from expyfun.io._parse import reconstruct_tracker, reconstruct_dealer
```
maddycapp27: Is this cheating? To import from `expyfun.io._parse` just seems a little sketchy.

expyfun/io/_parse.py
```diff
@@ -73,3 +99,102 @@ def read_tab(fname, group_start='trial_id', group_end='trial_ok'):
...
+        stopped state (as long as the trackers were allowed to stop during
+        the generation of the file.)
+    """
+    from ..stimuli import TrackerUD, TrackerBinom
+    # read in raw data
+    raw = read_tab_raw(fname)
+
+    # find tracker_identify and make list of IDs
+    tracker_idx = np.where([r[1] == 'tracker_identify' for r in raw])[0]
+    tr = []
```
maddycapp27: The way I have this currently set up is that any tracker object the function finds gets put in a list. Helpful when there are multiple tracker but vaguely annoying when there is only one and you want to do something like `tracker.plot()` but have to do `tracker[0].plot` instead. Thoughts?

expyfun/io/tests/test_parse.py
Eric89GXL: modify `expyfun/io/__init__.py` to import `reconstruct_tracker` and `reconstruct_dealer`, then you can just do `from expyfun.io import reconstruct_tracker, reconstruct_dealer`

Also don't nest these imports, put them at the top

expyfun/io/_parse.py
Eric89GXL: `tracker[0].plot` is fine, just be clear that a list is always returned (it can be length 0, 1, or more than 1 and the user should know to check)

expyfun/io/tests/test_parse.py
maddycapp27: Ah! Thank you!

maddycapp27:
@Eric89GXL @rkmaddox 

I think all the parts of this are in place and I'm open to suggestions on how to make it a little cleaner if you'd be so kind as to review the three new functions. Thanks much~

expyfun/io/tests/test_parse.py
```diff
@@ -45,3 +47,59 @@ def test_parse():
     data = read_tab(ec.data_fname, group_end=None)
     assert_equal(len(data[0]['misc']), 2)  # includes between-trials stuff
     assert_equal(len(data[1]['misc']), 2)
+
+
+def test_reconstruct():
+    """Test Tracker objects reconstruction"""
+
+    # test with one TrackerUD
+    with ExperimentController(*std_args, stim_fs=44100, **std_kwargs) as ec:
```
rkmaddox: Does ``stim_fs`` need to be specified for these tests?

expyfun/io/_parse.py
```diff
@@ -73,3 +99,110 @@ def read_tab(fname, group_start='trial_id', group_end='trial_ok'):
...
+            tr.append(TrackerBinom(**tracker_dict))
+        tr[-1]._tracker_id = tracker_id  # make sure tracker has original ID
+        stop_str = 'tracker_' + str(tracker_id) + '_stop'
+        tracker_stop_idx = np.where([r[1] == stop_str for r in raw])[0]
+        if len(tracker_stop_idx) == 0:
+            raise ValueError('Tracker {} has not stopped. All Trackers '
+                             'must be stopped.'.format(tracker_id))
+        responses = json.loads(raw[tracker_stop_idx[0]][2])['responses']
+        # feed in responses from tracker_ID_stop
+        [tr[-1].respond(r) for r in responses]
```
rkmaddox: Since nothing is done with the list, this would be cleaner as:
```
for r in responses:
    tr[-1].respond(r)
```

expyfun/io/_parse.py
```diff
@@ -73,3 +99,110 @@ def read_tab(fname, group_start='trial_id', group_end='trial_ok'):
...
+            raise ValueError('TrackerDealer {} has not stopped. All dealers '
+                             'must be stopped.'.format(dealer_id))
+        dealer_stop_log = json.loads(raw[dealer_stop_idx[0]][2])
+        log_response_history = dealer_stop_log['response_history']
+        log_x_history = dealer_stop_log['x_history']
+        log_tracker_history = dealer_stop_log['tracker_history']
+
+        dealer[-1]._response_history = log_response_history
+        dealer[-1]._x_history = log_x_history
+        dealer[-1]._tracker_history = log_tracker_history
```
rkmaddox: also: ``dealer[-1]._stopped = True``

expyfun/io/_parse.py
```diff
@@ -73,3 +99,110 @@ def read_tab(fname, group_start='trial_id', group_end='trial_ok'):
...
+        dealer_init_str = 'dealer_' + str(dealer_id) + '_init'
+        dealer_dict_idx = np.where([r[1] == dealer_init_str
+                                    for r in raw])[0][0]
+        dealer_dict = ast.literal_eval(raw[dealer_dict_idx][2])
+        dealer_trackers = dealer_dict['trackers']
+
+        # match up tracker objects to id
+        trackers = reconstruct_tracker(fname)
+        tr_objects = []
+        for t in dealer_trackers:
```
rkmaddox: Is the dealer shape being set to match its original shape anywhere?

expyfun/io/tests/test_parse.py
```diff
@@ -45,3 +47,59 @@ def test_parse():
...
+    tracker = reconstruct_tracker(ec.data_fname)[0]
+    assert(tracker.stopped)
+    tracker.x_current
+
+    # tracker not stopped
+    with ExperimentController(*std_args, stim_fs=44100, **std_kwargs) as ec:
+        tr = TrackerUD(ec, 1, 1, 3, 1, 5, np.inf, 3)
+        tr.respond(np.random.rand() < tr.x_current)
+        assert(not tr.stopped)
+    assert_raises(ValueError, reconstruct_tracker, ec.data_fname)
```
rkmaddox: @Eric89GXL Should we be deleting the data file when we're done with it?

expyfun/io/tests/test_parse.py
maddycapp27: I copied and pasted from `test_parse()`, but upon further (initial) thought, I don't see why it should be.

maddycapp27: I can try taking it out and see if anything gets terribly upset.

expyfun/io/tests/test_parse.py
```diff
@@ -45,3 +47,59 @@ def test_parse():
...
+    with ExperimentController(*std_args, **std_kwargs) as ec:
+        ec.identify_trial(ec_id='one', ttl_id=[0])
+        ec.start_stimulus()
+        ec.write_data_line('misc', 'trial one')
+        ec.stop()
+        ec.trial_ok()
+        ec.write_data_line('misc', 'end')
+
+    assert_raises(ValueError, reconstruct_tracker, ec.data_fname)
+    assert_raises(ValueError, reconstruct_dealer, ec.data_fname)
```
Eric89GXL: For whoever asked, no you don't need to clean up because `_TempDir()` should auto-delete itself when garbage-collected by Python

expyfun/io/_parse.py
```diff
@@ -73,3 +99,117 @@ def read_tab(fname, group_start='trial_id', group_end='trial_ok'):
...
+        raise ValueError('There are no Trackers in this file.')
+    tr = []
+    for ii in tracker_idx:
+        tracker_id = ast.literal_eval(raw[ii][2])['tracker_id']
+        tracker_type = ast.literal_eval(raw[ii][2])['tracker_type']
+        # find tracker_ID_init lines and get dict
+        init_str = 'tracker_' + str(tracker_id) + '_init'
+        tracker_dict_idx = np.where([r[1] == init_str for r in raw])[0][0]
+        tracker_dict = json.loads(raw[tracker_dict_idx][2])
+        if tracker_type == 'TrackerUD':
```
drammock: trying to understand this "trick"...  doesn't your suggested redefinition of `tracker_dict` overwrite the `json.loads` definition?  Did you mean to say something like 
```python
tracker_dict = json.loads(raw[tracker_dict_idx][2])  # unchanged
td = dict(TrackerUD=TrackerUD, TrackerBinom=TrackerBinom)  # changed var name
tr.append(td[tracker_type](**tracker_dict))
```

Eric89GXL: Oh sorry, didn't notice `tracker_dict` was taken -- swap in a different variable name into my code like `tracker_classes` or something and it should be fine

expyfun/io/_parse.py
maddycapp27: now it is

maddycapp27:
Been meaning to ask: 

There's a lot of `ast.literal_eval()` interspersed with `json.loads()` (I started replacing with `json.loads()` when `ast.literal_eval()` wasn't working.) Is there an advantage to using the `json` function (other than it working)?

expyfun/io/_parse.py
rkmaddox: And are the trackers being stored in that shape within the dealer? The ``numpy.array`` of trackers needs to be ``reshape``d to that size.

expyfun/io/tests/test_parse.py
rkmaddox: Ah, didn't realize it was going into a temp directory. Cool.

expyfun/io/_parse.py
```diff
@@ -73,3 +99,117 @@ def read_tab(fname, group_start='trial_id', group_end='trial_ok'):
...
+
+        # force input responses/log data
+        dealer_stop_str = 'dealer_' + str(dealer_id) + '_stop'
+        dealer_stop_idx = np.where([r[1] == dealer_stop_str for r in raw])[0]
+        if len(dealer_stop_idx) == 0:
+            raise ValueError('TrackerDealer {} has not stopped. All dealers '
+                             'must be stopped.'.format(dealer_id))
+        dealer_stop_log = json.loads(raw[dealer_stop_idx[0]][2])
+
+        shape = dealer_dict['shape']
```
rkmaddox: ``dealer[-1]._trackers.reshape(shape)`` should do it I think.

rkmaddox: (down in line 211 makes the most sense probably)

maddycapp27: That was my instinct as well

maddycapp27:
@drammock @Eric89GXL All tests except one are passing--looks like that one Travis test that is finicky is finding some random error and coveralls seems happy enough. Ready for review--thanks much!

expyfun/io/_parse.py
```diff
@@ -4,6 +4,40 @@
...
+    assert (len(lines[0]) == 1 and lines[0][0][0] == '#')
+    #metadata = ast.literal_eval(lines[0][0][2:])
+    assert lines[1] == ['timestamp', 'event', 'value']
+    lines = lines[2:]
+
+    times = [float(line[0]) for line in lines]
+    keys = [line[1] for line in lines]
+    vals = [line[2] for line in lines]
+    idx = np.arange(len(lines))
+    data = [(times[ii], keys[ii], vals[ii]) for ii in idx]
```
Eric89GXL: Perhaps `list(zip(times, keys, vals))`?

expyfun/io/_parse.py
```diff
@@ -4,6 +4,40 @@
...
+    Parameters
+    ----------
+    fname : str
+        Input filename.
+
+    Returns
+    -------
+    data : list of tuple
+        The data with each line from the tab file being a tuple in a list.
+        Each tuple is of the form (``timestamp``, ``key``, ``value).
```
Eric89GXL: missing double-backticks on `value`

expyfun/io/_parse.py
```diff
@@ -4,6 +4,40 @@
 
 import numpy as np
 import csv
+import ast
+import json
+
+
+def read_tab_raw(fname):
+    """Read .tab file from expyfun output without segmenting into trials
```
Eric89GXL: period at end

expyfun/io/_parse.py
```diff
@@ -73,3 +99,118 @@ def read_tab(fname, group_start='trial_id', group_end='trial_ok'):
...
+                             'must be stopped.'.format(dealer_id))
+        dealer_stop_log = json.loads(raw[dealer_stop_idx[0]][2])
+
+        shape = dealer_dict['shape']
+        log_response_history = dealer_stop_log['response_history']
+        log_x_history = dealer_stop_log['x_history']
+        log_tracker_history = dealer_stop_log['tracker_history']
+
+        dealer[-1]._shape = shape
+        dealer[-1]._trackers.reshape(shape)
```
Eric89GXL: Is this shape tested anywhere? I don't think it operates inplace:

https://docs.scipy.org/doc/numpy-1.12.0/reference/generated/numpy.ndarray.reshape.html

You probably want `dealer[-1]._trackers.shape = shape`.

expyfun/io/tests/test_parse.py
```diff
@@ -45,3 +47,59 @@ def test_parse():
...
+    """Test Tracker objects reconstruction"""
+
+    # test with one TrackerUD
+    with ExperimentController(*std_args, **std_kwargs) as ec:
+        tr = TrackerUD(ec, 1, 1, 3, 1, 5, np.inf, 3)
+        while not tr.stopped:
+            tr.respond(np.random.rand() < tr.x_current)
+
+    tracker = reconstruct_tracker(ec.data_fname)[0]
+    assert(tracker.stopped)
```
Eric89GXL: use `assert_true`, `assert` statements can be skipped by `-O`

expyfun/io/_parse.py
rkmaddox: Whoa I did not know you could to that. Awesome.

expyfun/io/_parse.py
rkmaddox: I don't think we do periods on the summaries, do we?

expyfun/io/_parse.py
Eric89GXL: You can only do it on appropriately contiguous arrays (i.e., won't require a copy), e.g.:
```
>>> import numpy as np
>>> x = np.eye(3).T
>>> x.shape = (9,)
  File "<stdin>", line 1, in <module>
AttributeError: incompatible shape for a non-contiguous array
```
But if you start with something effectively `ravel()`ed like here, then it should be safe.

expyfun/io/_parse.py
Eric89GXL: PEP257 says we should. I noticed that some/most of the other functions here had them so I figured we might as well.

rkmaddox: Huh. Maybe a good intro to git for the next person who needs to learn it.

maddycapp27:
@Eric89GXL In regards to examples, I'm not sure there's much to add--were you thinking of adding to the "parsing demo" or modifying one of the tracker examples (having them print to a .tab file instead of the console and then just running the plots from the reconstruction)?

maddycapp27:
Went ahead and put it in the ``TrackerDealer`` example--seemed the simplest option

examples/experiments/tracker_dealer.py
```diff
@@ -1,11 +1,13 @@
...
-======================
+==========================================================================
+Adaptive tracking for two trial types and tracker reconstruction from .tab
+==========================================================================
 
 This file shows how to interleave multiple Tracker objects using
-:class:`expyfun.stimuli.TrackerDealer`.
+:class:`expyfun.stimuli.TrackerDealer` as well as how to reconstruct the
+dealer from the .tab file logged by experiment controller with the
+``reconstruct_dealer()`` function.
```
Eric89GXL:
```
:func:`expyfun.io.reconstruct_dealer`
```

examples/experiments/tracker_dealer.py
```diff
@@ -45,48 +49,62 @@
...
-
-# Initialize human state
-rng_human = np.random.RandomState(1)  # random seed for modeled subject
-
-for ss, level in td:
-    # Get information of which trial type is next and what the level is at
-    # that time from TrackerDealer
-    td.respond(rng_human.rand() < sigmoid(level - true_thresh[sum(ss)],
-                                          lower=chance, slope=slope))
+# for that trial can be acquired. ExperimentController is used to generate log
```
Eric89GXL: Add `:class:` links here please

examples/experiments/tracker_dealer.py
```diff
@@ -45,48 +49,62 @@
...
+        # that time from TrackerDealer
+        td.respond(rng_human.rand() < sigmoid(level - true_thresh[sum(ss)],
+                                              lower=chance, slope=slope))
+
+###############################################################################
+# Reconstructing the TrackerDealer Object
+# ---------------------------------------
+# The TrackerDealer object has many built in analysis functions that are can
+# only be access through the object itself (not the log files alone). By using
+# ``reconstruct_dealer()``, the object can be recreated such that the analysis
```
Eric89GXL: `:func:` links

expyfun/io/_parse.py
```diff
@@ -4,10 +4,43 @@
...
+    Parameters
+    ----------
+    fname : str
+        Input filename.
+
+    Returns
+    -------
+    data : list of tuple
+        The data with each line from the tab file being a tuple in a list.
+        Each tuple is of the form (``timestamp``, ``key``, ``value``).
```
Eric89GXL: `See Also` section with `read_tab` (and add the reverse to `read_tab` docstring)

expyfun/io/__init__.py
```diff
@@ -2,7 +2,8 @@
 from ._wav import read_wav, write_wav
 from .._externals._h5io import (read_hdf5 as _read_hdf5,
                                 write_hdf5 as _write_hdf5)
-from ._parse import read_tab
+from ._parse import (read_tab, reconstruct_tracker, 
+                     reconstruct_dealer, read_tab_raw)
```
Eric89GXL: also need to add `read_tab_raw` to `doc/python_reference.rst`

expyfun/io/tests/test_parse.py
```diff
@@ -45,3 +47,61 @@ def test_parse():
     data = read_tab(ec.data_fname, group_end=None)
     assert_equal(len(data[0]['misc']), 2)  # includes between-trials stuff
     assert_equal(len(data[1]['misc']), 2)
+
+
+def test_reconstruct():
```
Eric89GXL: `_hide_window` decorator

expyfun/io/tests/test_parse.py
```diff
@@ -45,3 +47,61 @@ def test_parse():
...
+    tracker.x_current
+
+    # test with one TrackerBinom
+    with ExperimentController(*std_args, **std_kwargs) as ec:
+        tr = TrackerBinom(ec, .05, .5, 10)
+        while not tr.stopped:
+            tr.respond(True)
+
+    tracker = reconstruct_tracker(ec.data_fname)[0]
+    assert(tracker.stopped)
```
Eric89GXL: `assert_true`

examples/experiments/tracker_dealer.py
```diff
@@ -45,48 +49,63 @@
...
-
-# Initialize human state
-rng_human = np.random.RandomState(1)  # random seed for modeled subject
-
-for ss, level in td:
-    # Get information of which trial type is next and what the level is at
-    # that time from TrackerDealer
-    td.respond(rng_human.rand() < sigmoid(level - true_thresh[sum(ss)],
-                                          lower=chance, slope=slope))
+# for that trial can be acquired. :class:`ExperimentController` is used to
```
Eric89GXL: `expyfun.ExperimentController`

rkmaddox:
+1 merge after tests pass. I'm about to talk in a session at ASA, so I might not check in on the PR until this afternoon. But I'm happy if anyone pushes the merge button.

PR #318, opened 2017-06-26
WIP: Eyelink origin discrepancy

maddycapp27:
(copied from issue description) After setting up the eyelink I noticed that there was a discrepancy between coordinate systems.

`el.get_eye_position()` returns location in pixels with the center of the screen being `[0, 0]`. The eyelink itself claims to output a location where the top left is `[0, 0]` (but it seems more like it's using the bottom left). `ec._convert_units()`, however, assumes that the pixel locations use the bottom left of the screen as `[0, 0]`.

Since `el.wait_for_fix()` uses both `el.get_eye_position()` and `ec._convert_units()`, the coordinates never match up.

### DIFF
--- a/expyfun/_eyelink_controller.py
+++ b/expyfun/_eyelink_controller.py
@@ -603,7 +603,6 @@ class EyelinkController(object):
                 eye_pos = np.array(sample.getRightEye().getGaze())
             else:
                 eye_pos = np.array([np.inf, np.inf])
-            eye_pos -= (self._size / 2.)
         else:
             # use mouse, already referenced to center
             eye_pos = self._ec.get_mouse_position()


maddycapp27:
@Eric89GXL @drammock @rkmaddox 

Versions:
EyeLink 1000 Plus
PyLink 0.3.3 _maybe_ ... long story short, we are using a version of pylink that was packaged in with the EyeLink software because it was the only version that had something called `EyeLinkCustomDisplay` (which was a whole confusion in and of itself.

Eric89GXL:
So you're on Windows and you got it from here?

https://www.sr-support.com/forum/downloads/eyelink-display-software/39-eyelink-developers-kit-for-windows-windows-display-software

Eric89GXL:
Nice, one-line (-1) PR. I hope it ends up being correct because that's awesome :)

Does this fix things for you? Are you using that sample experiment to test? If so then we just need to wait until we can test it at UW.

Also, were calibrations affected at all by this bug at your end? I doubt it because I don't think it actually uses the position-fetching function, but want to double-check.

maddycapp27:
> So you're on Windows and you got it from here?

@rkmaddox I don't think I was here when you installed the software initially. But this is what's described in the manual so I assume that's what you and Kevin did?

> Does this fix things for you? Are you using that sample experiment to test? 

Yes to both.

> Also, were calibrations affected at all by this bug at your end?

Nope, that was part of why I was so befuddled when I was troubleshooting.

rkmaddox:
Nope, got the SDK from the USB stick that came with the Eyelink. Same as the ``pylink`` software (which looked completely different than ``pylink`` on github--how is that possible?)

Worth noting that our Eyelink computer is definitely running different software. It has a GUI DOS interface. Don't know if that is where the origin difference is coming from though.

Eric89GXL:
The EL acquisition system at UW has a DOS GUI, too, so I'm not sure what
you mean about them being different.

rkmaddox:
Sorry, I wasn't clear. When you exit the tracker program at UW, I seem to recall it drops to a DOS prompt (and you just run ``t`` to start it again). When you exit it at UR, it drops to a system GUI.

Eric89GXL:
That's my understanding of how it works at UW, too.

So the versions you think you have are EyeLink 1000 Plus, PyLink 0.3.3. Is there any OS information in the GUIs (either your "OS" gui or the Eyelink calibration/run GUI)? Might as well get everything documented in case cross-site problems crop up.

maddycapp27:
I'll check on that first thing tomorrow morning!

maddycapp27:
>  Is there any OS information in the GUIs (either your "OS" gui or the Eyelink calibration/run GUI)? 

It says Version 5.09.

Eric89GXL:
UW:
EyeLink 1000 v 4.56 complied 18 Aug 2010
pylink.version.vernum `(1, 0, 0, 119)`
Pyglet 1.2.4
Found an installer for "EyeLinkDevKit_1.9.214" FWIW

It looks like this doesn't work with whatever (very old) `master` was on the machine but does work on this PR. So I think this indeed fixes a bug.

@maddycapp27 is this just WIP because you were waiting for us to check?

maddycapp27:
> is this just WIP because you were waiting for us to check?

Yes, sorry for confusion.

Eric89GXL:
Thanks @maddycapp27

maddycapp27:
Thank you @Eric89GXL

PR #319, opened 2017-07-10
MRG: CRM Corpus module (part 2)

rkmaddox:
Adds functionality for preparing a CRM corpus at a certain sampling rate, and then provides methods for using it, either loading files from hard drive one by one, or by loading all sounds into memory first for fast access.

Continues #247.

Still need to add tests, but would like comments on what is there so far.

### DIFF
--- a/examples/stimuli/crm_stimuli.py
+++ b/examples/stimuli/crm_stimuli.py
@@ -59,7 +59,7 @@ x = add_pad([x1, x2], alignment='start')
 ###############################################################################
 # Now we actually run the experiment.
 
-max_wait = 0 if building_doc else 3
+max_wait = 0.01 if building_doc else 3
 with ExperimentController(
         exp_name='CRM corpus example', window_size=(720, 480),
         full_screen=False, participant='foo', session='foo', version='dev',
--- a/examples/stimuli/crm_stimuli.py
+++ b/examples/stimuli/crm_stimuli.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+"""
+========================
+Generate vocoded stimuli
+========================
+
+This shows how to use the CRM corpus functions.
+
+@author: rkmaddox
+"""
+
+from expyfun._utils import _TempDir
+from expyfun import ExperimentController, analyze
+from expyfun.stimuli import (crm_prepare_corpus, crm_sentence, crm_info,
+                             crm_response_menu, add_pad, CRMPreload)
+
+print(__doc__)
+
+crm_path = _TempDir()
+fs = 40000
+ec_args = dict()
+
+# prepare the corpus for two talkers at the native 40000 Hz sampling rate
+# NOTE: For your experiment, you only need to prepare the corpus once per
+# sampling rate, you should probably use the default path, and you should just
+# do all the talkers at once. For the example, we are using fs=40000 and only
+# doing two talkers so that the stimulus preparation is very fast, and a temp
+# directory so that we don't interfere with any other prepared corpuses. Your
+# code will likely look like this commented line, and not appear in your
+# actual experiment script:
+# crm_prepare_corpus(fs)  # where fs is likely 24414 or 44100
+crm_prepare_corpus(fs, path_out=crm_path, overwrite=True,
+                   talker_list=[dict(sex=0, talker_num=0),
+                                dict(sex=1, talker_num=0)])
+
+# print the valid callsigns
+print('Valid callsigns are {0}'.format(crm_info()['callsign']))
+
+# read a sentence in from the hard drive
+x1 = 0.5 * crm_sentence(fs, 'm', '0', 'c', 'r', '5', path=crm_path)
+
+# preload all the talkers and get a second sentence
+crm = CRMPreload(fs, path=crm_path)
+x2 = crm.sentence('f', '0', 'ringo', 'green', '6')
+
+x = add_pad([x1, x2], alignment='start')
+
+with ExperimentController(
+        exp_name='CRM corpus example', window_size=(720, 480),
+        full_screen=False, participant='foo', session='foo', version='dev',
+        output_dir=None, stim_fs=40000) as ec:
+    ec.screen_text('Report the color and number spoken by the female '
+                   'talker.', wrap=True)
+    screenshot = ec.screenshot()
+    ec.flip()
+    ec.wait_secs(3)
+
+    ec.load_buffer(x)
+    ec.identify_trial(ec_id='', ttl_id=[])
+    ec.start_stimulus()
+    ec.wait_secs(x.shape[-1] / float(fs))
+
+    resp = crm_response_menu(ec)
+    if resp == ('g', '6'):
+        ec.screen_prompt('Correct!', max_wait=1, min_wait=1)
+    else:
+        ec.screen_prompt('Incorrect.', max_wait=1, min_wait=1)
+    ec.trial_ok()
+
+analyze.plot_screen(screenshot)
--- a/examples/stimuli/crm_stimuli.py
+++ b/examples/stimuli/crm_stimuli.py
@@ -59,6 +59,7 @@ x = add_pad([x1, x2], alignment='start')
 ###############################################################################
 # Now we actually run the experiment.
 
+max_wait = 0.01 if building_doc else 3
 with ExperimentController(
         exp_name='CRM corpus example', window_size=(720, 480),
         full_screen=False, participant='foo', session='foo', version='dev',
@@ -67,7 +68,7 @@ with ExperimentController(
                    'talker.', wrap=True)
     screenshot = ec.screenshot()
     ec.flip()
-    ec.wait_secs(3)
+    ec.wait_secs(max_wait)
 
     ec.load_buffer(x)
     ec.identify_trial(ec_id='', ttl_id=[])
@@ -76,9 +77,9 @@ with ExperimentController(
 
     resp = crm_response_menu(ec, max_wait=0.01 if building_doc else np.inf)
     if resp == ('g', '6'):
-        ec.screen_prompt('Correct!', max_wait=3, min_wait=1)
+        ec.screen_prompt('Correct!', max_wait=max_wait)
     else:
-        ec.screen_prompt('Incorrect.', max_wait=3, min_wait=1)
+        ec.screen_prompt('Incorrect.', max_wait=max_wait)
     ec.trial_ok()
 
 analyze.plot_screen(screenshot)
--- a/examples/stimuli/crm_stimuli.py
+++ b/examples/stimuli/crm_stimuli.py
@@ -75,7 +75,7 @@ with ExperimentController(
     ec.start_stimulus()
     ec.wait_secs(x.shape[-1] / float(fs))
 
-    resp = crm_response_menu(ec, max_wait=0 if building_doc else np.inf)
+    resp = crm_response_menu(ec, max_wait=0.01 if building_doc else np.inf)
     if resp == ('g', '6'):
         ec.screen_prompt('Correct!', max_wait=max_wait, min_wait=1)
     else:
--- a/examples/stimuli/crm_stimuli.py
+++ b/examples/stimuli/crm_stimuli.py
@@ -59,7 +59,6 @@ x = add_pad([x1, x2], alignment='start')
 ###############################################################################
 # Now we actually run the experiment.
 
-max_wait = 0.01 if building_doc else 3
 with ExperimentController(
         exp_name='CRM corpus example', window_size=(720, 480),
         full_screen=False, participant='foo', session='foo', version='dev',
@@ -68,7 +67,7 @@ with ExperimentController(
                    'talker.', wrap=True)
     screenshot = ec.screenshot()
     ec.flip()
-    ec.wait_secs(max_wait)
+    ec.wait_secs(3)
 
     ec.load_buffer(x)
     ec.identify_trial(ec_id='', ttl_id=[])
@@ -77,9 +76,9 @@ with ExperimentController(
 
     resp = crm_response_menu(ec, max_wait=0.01 if building_doc else np.inf)
     if resp == ('g', '6'):
-        ec.screen_prompt('Correct!', max_wait=max_wait, min_wait=1)
+        ec.screen_prompt('Correct!', max_wait=3, min_wait=1)
     else:
-        ec.screen_prompt('Incorrect.', max_wait=max_wait, min_wait=1)
+        ec.screen_prompt('Incorrect.', max_wait=3, min_wait=1)
     ec.trial_ok()
 
 analyze.plot_screen(screenshot)
--- a/examples/stimuli/crm_stimuli.py
+++ b/examples/stimuli/crm_stimuli.py
@@ -10,9 +10,11 @@ This shows how to use the CRM corpus functions.
 """
 
 from expyfun._utils import _TempDir
-from expyfun import ExperimentController, analyze
+from expyfun import ExperimentController, analyze, building_doc
 from expyfun.stimuli import (crm_prepare_corpus, crm_sentence, crm_info,
                              crm_response_menu, add_pad, CRMPreload)
+
+import numpy as np
 
 print(__doc__)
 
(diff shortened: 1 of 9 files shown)


rkmaddox:
Will add an example, too, once everyone is happy. Tests will be tricky since it takes a long time to prepare the corpus, I just realized.

rkmaddox:
Sphinx build seems to fail due to use of `joblib`. I see now that that's an optional requirement. I should condition that import, huh?

Eric89GXL:
use `._parallel` functions

Eric89GXL:
Regarding testing, is it possible to specify some small subset of stimuli to generate? To avoid exposing it publicly, we could break the real functionality into some private function that accepts such a parameter.

rkmaddox:
For testing, yeah that would be possible. I had (but removed because it's messy) the option of preparing a subset of talkers, so all it would require is putting that argument back in.

But, it still takes about a minute to do one talker using 11 cores on my machine, which really sucks for the tests. I could see if resampling to 20k (simple 0.5 downsample) runs faster.

Eric89GXL:
>  I could see if resampling to 20k (simple 0.5 downsample) runs faster.

Or maybe no resampling?

rkmaddox:
Neither 0.5x or 1x resampling sped it up. However, If I condition the resampling thusly:

```
if int(np.round(fs_out)) != int(np.round(_fs_binary)):
        x = resample(x, fs_out, _fs_binary, n_jobs=n_jobs, verbose=0)
```
Then it takes about 30 seconds to prepare the whole corpus.

Eric89GXL:
I think the special case + single-talker is worth it for testing

rkmaddox:
Wilco.

expyfun/stimuli/_crm.py
```diff
@@ -0,0 +1,487 @@
...
+    '7': 6,
+    '8': 7,
+    0: 0,
+    1: 1,
+    2: 2,
+    3: 3,
+    4: 4,
+    5: 5,
+    6: 6,
+    7: 7}
```
drammock: I'm sure you have a good reason for doing it like this, but man, does this look scary:
```python
numbers['1'] == 0
numbers[1] == 1
```

rkmaddox: Yeah, there's a warning about it elsewhere in the code. Basically, I wanted to give people access to the corpus via the "name" of each parameter ('charlie', '1'), or by its index in the list of options. And the index of '1' is 0.

Eric89GXL: Perhaps you can also do some stuff like `{str(x + 1): x for x in range(10)}` to make it a bit more readable

drammock: @Eric89GXL I'm not sure I agree that it would be more readable that way for all users.  And in any case it wouldn't work for the spelled-out versions `'one'`, `'two'` etc. so I'm inclined to leave it as-is, and just add an extra comment here that points source-code-reading-users to the part of the docs that explains the scheme more thoroughly.

rkmaddox:
OK. I beefed up the warnings about the number indexing stuff. I also made the changes that allow you to prepare just a subset of talkers (which was no fun). Other than a bunch of tests :-1: :-1: :-1:, I think it's ready to go. Any other comments?

rkmaddox:
Ah crap, gotta do the parallel stuff.

rkmaddox:
Hoping tests pass. @larsoner Is there a reason why coveralls has only run on one of the many recent commits?

larsoner:
It looks like it ran on the one where all Travis runs passed

rkmaddox:
Got it. Here's hoping that happens with this one, too.

rkmaddox:
Getting weird HTTP errors in AppVeyor. I though it might be a windows \\ vs / thing, so I added something to fix that (which we probably need anyway), but it looks like those are using Arch Linux? Any ideas why `fetch_data_file` might not work there?

expyfun/stimuli/_crm.py
```diff
@@ -0,0 +1,523 @@
...
+    if value in param_dict.keys():
+        return param_dict[value]
+    else:
+        raise ValueError('{} is not a valid {}. Legal values are: {}'
+                         .format(value, name, sorted(param_dict.keys())))
+
+
+def _read_talker_zip_file(sex, talker_num):
+    talker_num_raw = _n_talkers * _sexes[sex] + _talker_nums[talker_num]
+    fn = fetch_data_file(join('crm', 'Talker%i.zip' % talker_num_raw))
```
larsoner: don't use `join` use `/` directly because the input is really a URL (then you don't need your replace above)

rkmaddox: OK. But won't that write to weird locations in Windows?

larsoner: You would think so but actually the Python interpreter takes `/` as directory even on windows (so using `op.join` anywhere might even be overkill in the end)

larsoner:
>  but it looks like those are using Arch Linux?

What do you mean? AppVeyor should run Windows, Travis and CircleCI run Ubuntu

expyfun/stimuli/_crm.py
rkmaddox: Huh. OK.

rkmaddox:
OK, that makes more sense.

Ahhhh `ARCH` is for architecture. Not Arch. Got it.

expyfun/stimuli/_crm.py
```diff
@@ -132,7 +132,7 @@ def _check(name, value):
 
 def _read_talker_zip_file(sex, talker_num):
     talker_num_raw = _n_talkers * _sexes[sex] + _talker_nums[talker_num]
-    fn = fetch_data_file(join('crm', 'Talker%i.zip' % talker_num_raw))
+    fn = 'crm/Talker%i.zip' % talker_num_raw
```
larsoner: did you mean to delete the `fetch_data_file` call?

rkmaddox: Ah crap.

expyfun/stimuli/_stimuli.py
```diff
@@ -131,3 +131,57 @@ def play_sound(sound, fs=None, norm=True, wait=False):
...
+        number of channels (by appending a copy), and the length will be equal
+        to the maximum length (by appending zeros).
+
+    Notes
+    -----
+        Even if the original sounds were all 0- or 1-dimensional, the output
+        will be 2-dimensional (channels, samples).
+    """
+    if alignment not in ['start', 'center', 'end']:
+        raise(ValueError("alignment must be either 'start', 'center', "
```
larsoner: `raise ValueError(...)`

expyfun/stimuli/_stimuli.py
```diff
@@ -131,3 +131,57 @@ def play_sound(sound, fs=None, norm=True, wait=False):
...
+            elif alignment == 'end':
+                n_pre = len_max - length
+                n_post = 0
+            x[xi] = np.concatenate((np.zeros((ch, n_pre)),
+                                    x[xi],
+                                    np.zeros((ch, n_post))),
+                                   axis=1)
+        if ch < ch_max:
+            x[xi] = np.tile(x[xi], [ch_max, 1])
+    return np.array(x).sum(0)
```
larsoner: can be just `np.sum(...)`

expyfun/stimuli/_stimuli.py
```diff
@@ -131,3 +131,57 @@ def play_sound(sound, fs=None, norm=True, wait=False):
...
+    Notes
+    -----
+        Even if the original sounds were all 0- or 1-dimensional, the output
+        will be 2-dimensional (channels, samples).
+    """
+    if alignment not in ['start', 'center', 'end']:
+        raise(ValueError("alignment must be either 'start', 'center', "
+                         "or 'end'"))
+    x = [np.atleast_2d(y) for y in sounds]
+    if not np.all([y.ndim == 2 for y in x]):
```
larsoner: no need for `[` enclosing here

better `all(y.ndim == 2 for y in x)` as it's a generator expression that will short-circuit

expyfun/stimuli/_crm.py
```diff
@@ -0,0 +1,523 @@
...
+    for bi in np.arange(0, len(raw), 2, dtype=int):
+        x[bi / 2] = struct.unpack('<h', raw[bi:bi + 2])[0]
+    x /= 16384.
+    if ramp_dur:
+        return window_edges(x, _fs_binary, dur=ramp_dur)
+    else:
+        return x
+
+
+def _prepare_stim(zip_file, path_out, sex, tal, cal, col, num, fs_out, dtype,
```
larsoner: do not pass an open file handle to something that gets parallelized, you are asking for trouble (and Travis agrees) -- pass the filename

rkmaddox:
So it looks like things are bombing because they don't have ``joblib``. Should I put a ``try``/``except`` around my parallel stuff? It seems like the point of ``parallel_func`` was to use ``joblib`` if it was available, but to still run if it wasn't, but that's not what it does. Am I misunderstanding its use?

expyfun/stimuli/_stimuli.py
```diff
@@ -131,3 +131,57 @@ def play_sound(sound, fs=None, norm=True, wait=False):
...
+            elif alignment == 'center':
+                n_pre = (len_max - length) // 2
+                n_post = len_max - length - n_pre
+            elif alignment == 'end':
+                n_pre = len_max - length
+                n_post = 0
+            x[xi] = np.concatenate((np.zeros((ch, n_pre)),
+                                    x[xi],
+                                    np.zeros((ch, n_post))),
+                                   axis=1)
```
larsoner: can you see if `np.pad` would simplify this bit?

expyfun/stimuli/_crm.py
```diff
@@ -0,0 +1,523 @@
...
+import os
+from os.path import join
+import numpy as np
+from mne.filter import resample
+import expyfun.visual as vis
+from ._stimuli import window_edges
+from ..io import read_wav, write_wav
+from .._utils import fetch_data_file, _get_user_home_path
+from zipfile import ZipFile
+from expyfun._parallel import parallel_func
```
larsoner: all `expyfun` imports within the module should be relative, so `from ..parallel import parallel_func`

expyfun/stimuli/_crm.py
```diff
@@ -0,0 +1,523 @@
...
+"""
+
+# Author: Ross Maddox <ross.maddox@rochester.edu>
+#
+# License: BSD (3-clause)
+
+import os
+from os.path import join
+import numpy as np
+from mne.filter import resample
```
larsoner: nest this within whatever uses `resample`

expyfun/stimuli/_stimuli.py
rkmaddox: I gave it a shot, but ``np.pad`` didn't handle the axes properly.

larsoner: It sounds [pretty flexible](https://docs.scipy.org/doc/numpy/reference/generated/numpy.pad.html):
```
Number of values padded to the edges of each axis. ((before_1, after_1), ... (before_N, after_N)) unique pad widths for each axis. ((before, after),) yields same before and after pad for each axis.
```

rkmaddox: Got it. Fixed that.

expyfun/stimuli/_crm.py
rkmaddox: Since the zip file contains a bunch of files itself, this means that each zip file will have to be opened 256 times instead of once. Since it is reading different binary files from the zip file each time, is it a problem?

Also adds new errors where previously it was passing. I will try to track them down.

rkmaddox: NM on the error, I think I figured out what it was.

larsoner: It should be okay to reopen, but you should check to see if it slows things down.

AFAIK you cannot safely pass around an open file handle (and ZipFile class is/has one) and expect it to behave nicely in all threads. You have three options:

1. Read all data in at the start, then pass `ndarray` to parallel functions
2. Pass filename and open handles within the parallel function
3. Don't parellelize
4. Figure out how to duplicate the ZipFile handle in a parallel-safe way (probably ends up being equivalent to (2), but harder to work out)

rkmaddox:
Oh my God, all the tests passed. Ready for review and merge.

rkmaddox:
This lengthens the tests significantly, because it needs to download a talker zip file. I don't know a way around this...

larsoner:
> This lengthens the tests significantly, because it needs to download a talker zip file.

It's only 28 MB, is it really the bottleneck? The test takes ~1 minute

larsoner:
(for reference, in MNE each Travis build pulls roughly 1 GB of data, and this step takes [~40 sec](https://travis-ci.org/mne-tools/mne-python/jobs/263116203))

rkmaddox:
I guess a minute isn't too bad, but it's way longer than any of the other tests. If you're ok with it I am.

larsoner:
It suggests that some other step is slow, have you profiled at all? If not I can take a look tomorrow

rkmaddox:
Nope, I have not. Could also be the stimulus prep. Takes about 4 seconds on my computer using 11 cores.

rkmaddox:
After profiling: yep, it's the processing, not the download. On my computer it was 16 s of stimulus prep and 4 s of download. That difference is likely to be exaggerated on a VM whose download time is similar to mine, but whose CPUs are fewer and slower.

The stim prep is slow because it opens, copies, and closes 256 little sound files (256 per talker, doing one talker in the tests). It does it once using all the cores, and once using one core. It skips the resampling bit by staying at the native sampling rate of 40000.

I don't think there is much I can do to make it faster.

larsoner:
https://github.com/rkmaddox/expyfun/pull/3

rkmaddox:
Oh my god it's so much faster. How embarrassing for me. I cannot believe I left a loop through data in there! Thanks @larsoner.

larsoner:
Only thing remaining that I can see is some example exposing the functionality

rkmaddox:
Cool. I started the example today but didn't quite have time to finish. Will push tomorrow.

larsoner:
@rkmaddox let me know when you're happy, I'll take one last look and merge

rkmaddox:
Assuming tests pass I am very happy. Thanks for the extensive review so far.

larsoner:
Oh crap that example won't build properly because it actually waits for a response. You'll want to add a conditional `max_wait = 0. if building_doc else np.inf` and pass it to `crm_response_menu`

larsoner:
(using `from expyfun import building_doc`)

larsoner:
`ValueError: min_wait must be less than max_wait`

examples/stimuli/crm_stimuli.py
```diff
@@ -0,0 +1,85 @@
...
+# preload all the talkers and get a second sentence from memory
+crm = CRMPreload(fs, path=crm_path)
+x2 = crm.sentence('f', '0', 'ringo', 'green', '6')
+
+x = add_pad([x1, x2], alignment='start')
+
+###############################################################################
+# Now we actually run the experiment.
+
+max_wait = 0 if building_doc else 3
```
larsoner: this one, too...

examples/stimuli/crm_stimuli.py
```diff
@@ -0,0 +1,85 @@
...
+    ec.load_buffer(x)
+    ec.identify_trial(ec_id='', ttl_id=[])
+    ec.start_stimulus()
+    ec.wait_secs(x.shape[-1] / float(fs))
+
+    resp = crm_response_menu(ec, max_wait=0.01 if building_doc else np.inf)
+    if resp == ('g', '6'):
+        ec.screen_prompt('Correct!', max_wait=max_wait, min_wait=1)
+    else:
+        ec.screen_prompt('Incorrect.', max_wait=max_wait, min_wait=1)
```
larsoner: `min_wait` here is `1` so `max_wait` must be at least `1` (you can just always make it `3` for simplicity, it's okay if it adds 3 sec to the doc build)

larsoner:
[Victory!](https://257-11614571-gh.circle-artifacts.com/0/home/ubuntu/expyfun/doc/_build/html/auto_examples/stimuli/crm_stimuli.html#sphx-glr-auto-examples-stimuli-crm-stimuli-py)

larsoner:
AppVeyor failures are unrelated and Travis is just waiting on OSX (which can take forever) so I'll merge, thanks @rkmaddox

rkmaddox:
Woohoo!

PR #320, opened 2017-07-13
MRG: Fix version

Eric89GXL:
Our current version code returns junk like `.dev0+c` so version assertion round-trips fail. This fixes the problem (there was a lingering `[0]` in `setup.py`)

### DIFF
--- a/expyfun/tests/test_version.py
+++ b/expyfun/tests/test_version.py
@@ -2,6 +2,7 @@
 import os
 from os import path as op
 import warnings
+import shutil
 
 from nose.tools import assert_raises, assert_true, assert_equal
 
@@ -9,9 +10,6 @@ from expyfun import (ExperimentController, assert_version, download_version,
                      __version__)
 from expyfun._utils import _TempDir
 from expyfun._git import _has_git
-
-tempdir = _TempDir()
-tempdir_2 = _TempDir()
 
 
 def test_version():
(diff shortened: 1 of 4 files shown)


Eric89GXL:
CIs should be happy with latest commit, ready for review/merge from my end.

Eric89GXL:
ping @drammock, can't use `download_version` in master / finalize experiments without this. CI failures and just from me killing the `branch` builds.

expyfun/tests/test_version.py
```diff
@@ -24,35 +21,45 @@ def test_version():
...
-        download_version(dest_dir=tempdir_2)
+    for want_version in ('090948e', 'cae6bc3', 'b6e8a81'):  # old, broken, new
+        tempdir = _TempDir()
+        tempdir_2 = _TempDir()
+        if not _has_git:
+            assert_raises(ImportError, download_version, want_version, tempdir)
+        else:
+            assert_raises(IOError, download_version, want_version,
+                          op.join(tempdir, 'foo'))
+            assert_raises(RuntimeError, download_version, 'x' * 7, tempdir)
```
drammock: running `make test` locally fails here: `OSError: Destination directory /tmp/tmptkzwpcpa already has "expyfun" subdirectory`

drammock: I'm not sure how that's possible, since new `_TempDir()`s are getting created each time.  Can you confirm?

Eric89GXL: I do not get this error, not sure why you do. Are you sure you're up to date? Note this is on `upstream/fix-version` not `Eric89GXL/fix-version`

expyfun/tests/test_version.py
```diff
@@ -24,35 +21,45 @@ def test_version():
...
+            orig_dir = os.getcwd()
+            os.chdir(tempdir)
+            try:
+                assert_true(op.isdir('expyfun'))
+                assert_raises(IOError, download_version, want_version)
+            finally:
+                os.chdir(orig_dir)
+            # make sure we can get latest version
+            assert_raises(IOError, download_version, dest_dir=tempdir)
+            download_version(dest_dir=tempdir_2)
```
drammock: shouldn't these last two lines be outdented?  They're running with each iteration of the for loop, but don't reference `want_version` anywhere.

Eric89GXL: The first one is very fast so no harm in keeping it, but the second one yeah it would be faster to move it

expyfun/tests/test_version.py
drammock: Ran it twice, seems always to fail on the 3rd run of the for-loop.

drammock: oh, you're right.  I was testing `Eric89GXL/fix-version`.  Tests pass on `upstream/fix-version`

Eric89GXL:
Comments addressed in latest commit. Good to merge once CIs come back happy? For simplicity I won't cancel any :)

drammock:
+1 when CIs are happy.  Why is it running appveyor and travis twice (once as "pr" and once as "push")?  What's the difference?

Eric89GXL:
"push" and "branch" are both pushes of a branch to this repo, just different names

PR is PR into this repo

This PR does both whenever I push a commit

Eric89GXL:
Push re-build is taking too long but all others pass. Thanks for the quick look @drammock

PR #321, opened 2017-07-21
MRG: ProgressBar

maddycapp27:
``ProgressBar`` is a class to add to ``expyfun.visual`` that lets you display and update a progress bar during an experiment. I've been using it to let subjects know how far they are in an experiment during breaks between trials.

### DIFF
--- a/examples/experiments/progress_bar.py
+++ b/examples/experiments/progress_bar.py
@@ -45,11 +45,12 @@ with ExperimentController('name', version='dev', window_size=[500, 500],
                            ''.format(percent), [0, .1], wrap=False,
                            font_size=12)
             pb.draw()
+            if n == 4:
+                screenshot = ec.screenshot()
             ec.flip()
             # subject uses any key press to proceed
             ec.wait_one_press(max_wait=max_wait)
     ec.screen_text('This example is complete.')
-    screenshot = ec.screenshot()
     ec.flip()
     ec.wait_secs(1)
 
--- a/examples/experiments/progress_bar.py
+++ b/examples/experiments/progress_bar.py
@@ -8,7 +8,7 @@ ProgressBar demo
 This example shows how to display progress between trials using
 :class:`expyfun.visual.ProgressBar`.
 """
-from expyfun import ExperimentController
+from expyfun import ExperimentController, building_doc
 from expyfun.visual import ProgressBar
 import numpy as np
 
@@ -29,9 +29,11 @@ with ExperimentController('name', version='dev', window_size=[500, 500],
         number = np.random.randint(1, 5)
         ec.screen_text(str(number), wrap=False)
         ec.flip()
-        ec.wait_one_press(live_keys=[number])
+        if not building_doc:
+            ec.wait_one_press(live_keys=[number])
         ec.flip()
-        ec.wait_secs(0.5)
+        if not building_doc:
+            ec.wait_secs(0.5)
         # only show progress bar every other trial
         if n % 2 == 0:
             # calculate percent done and update the bar object
@@ -44,7 +46,8 @@ with ExperimentController('name', version='dev', window_size=[500, 500],
             pb.draw()
             ec.flip()
             # subject uses any key press to proceed
-            ec.wait_one_press()
+            if not building_doc:
+                ec.wait_one_press()
     ec.screen_text('This example is complete.')
     ec.flip()
     ec.wait_secs(1)
--- a/examples/experiments/progress_bar.py
+++ b/examples/experiments/progress_bar.py
@@ -5,7 +5,7 @@
 ProgressBar demo
 ================
 
-This example shows how to display progress between trials using 
+This example shows how to display progress between trials using
 :class:`expyfun.visual.ProgressBar`.
 """
 from expyfun import ExperimentController
@@ -17,13 +17,13 @@ n_trials = 6
 with ExperimentController('name', version='dev', window_size=[500, 500],
                           full_screen=False, session='foo',
                           participant='foo') as ec:
-    
+
     # initialize the progress bar
     pb = ProgressBar(ec, [0, -.1, 1.5, .1], units='norm')
 
     ec.screen_prompt('Press the number shown on the screen. Start by pressing'
                      ' 1.', font_size=12, live_keys=[1])
-    
+
     for n in np.arange(n_trials) + 1:
         # subject does some task
         number = np.random.randint(1, 5)
@@ -39,7 +39,7 @@ with ExperimentController('name', version='dev', window_size=[500, 500],
             pb.update_bar(percent)
             # display the progress bar with some text
             ec.screen_text('You\'ve completed {} %. Press any key to proceed.'
-                           ''.format(percent), [0, .1], wrap=False, 
+                           ''.format(percent), [0, .1], wrap=False,
                            font_size=12)
             pb.draw()
             ec.flip()
@@ -48,4 +48,3 @@ with ExperimentController('name', version='dev', window_size=[500, 500],
     ec.screen_text('This example is complete.')
     ec.flip()
     ec.wait_secs(1)
-        
--- a/examples/experiments/progress_bar.py
+++ b/examples/experiments/progress_bar.py
@@ -10,9 +10,12 @@ This example shows how to display progress between trials using
 """
 from expyfun import ExperimentController, building_doc
 from expyfun.visual import ProgressBar
+import expyfun.analyze as ea
 import numpy as np
 
 n_trials = 6
+max_wait = 0.1 if building_doc else np.inf
+wait_dur = 0.1 if building_doc else 0.5
 
 with ExperimentController('name', version='dev', window_size=[500, 500],
                           full_screen=False, session='foo',
@@ -22,18 +25,16 @@ with ExperimentController('name', version='dev', window_size=[500, 500],
     pb = ProgressBar(ec, [0, -.1, 1.5, .1], units='norm')
 
     ec.screen_prompt('Press the number shown on the screen. Start by pressing'
-                     ' 1.', font_size=12, live_keys=[1])
+                     ' 1.', font_size=12, live_keys=[1], max_wait=max_wait)
 
     for n in np.arange(n_trials) + 1:
         # subject does some task
         number = np.random.randint(1, 5)
         ec.screen_text(str(number), wrap=False)
         ec.flip()
-        if not building_doc:
-            ec.wait_one_press(live_keys=[number])
+        ec.wait_one_press(live_keys=[number], max_wait=max_wait)
         ec.flip()
-        if not building_doc:
-            ec.wait_secs(0.5)
+        ec.wait_secs(wait_dur)
         # only show progress bar every other trial
         if n % 2 == 0:
             # calculate percent done and update the bar object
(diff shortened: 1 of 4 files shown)


maddycapp27:
@drammock @Eric89GXL @rkmaddox 

First of all, no rush to look at this--it's just a quick side project.

I wasn't totally sure how much customization this called for, so let me know if I need to add (or remove I suppose) functionality in that regard. Otherwise, as always, I am generally open to feedback. Thanks much!

Eric89GXL:
I would expect outline color, fill color, and *maybe* a text display of the current count `N/N_total` -- not sure how that would fit aesthetically, though.

maddycapp27:
I'll think about the text display--it's worth a ponder to figure out how to make it look not clumsy.

expyfun/visual/_visual.py
```diff
@@ -826,6 +826,87 @@ def __init__(self, ec, colors=('w', 'k')):
...
+        self._pos_bar = pos.copy()
+        self._pos_bar[0] -= self._width * 0.5
+        self._init_x = self._pos_bar[0]
+        self._pos_bar[2] = 0
+
+        self._rectangles = [Rectangle(ec, self._pos_bar, units, colors[0],
+                                      None),
+                            Rectangle(ec, self._pos, units, None, colors[1])]
+        if units == 'norm':
+            self._text_pos = [pos[0], pos[1] - pos[3] * 0.5 - 0.1]
```
maddycapp27: I feel like there has to be a more eloquent way to do this. Especially for using units in pixels.

maddycapp27:
I'm giving up on making text work with this for the time being. Still waiting for the last Travis to finish and Appveyor failed on an unrelated thing but otherwise tests look ok. @larsoner @drammock @rkmaddox I'm ready for review. Thanks much!

expyfun/visual/_visual.py
```diff
@@ -826,6 +826,76 @@ def __init__(self, ec, colors=('w', 'k')):
...
+        self._pos_bar = pos.copy()
+        self._pos_bar[0] -= self._width * 0.5
+        self._init_x = self._pos_bar[0]
+        self._pos_bar[2] = 0
+
+        self._rectangles = [Rectangle(ec, self._pos_bar, units, colors[0],
+                                      None),
+                            Rectangle(ec, self._pos, units, None, colors[1])]
+
+    def update_bar(self, percent):
```
larsoner: percent is between 0 and 100, so either change the range or call it proportion

expyfun/visual/_visual.py
```diff
@@ -826,6 +826,76 @@ def __init__(self, ec, colors=('w', 'k')):
...
+        Colors to fill and outline the bar respectively. Defaults to green and
+        white.
+    text : bool
+        Whether to show the percent done as text below the bar. If ''True'',
+        leave room on the screen below the bar
+
+    Returns
+    -------
+    bar : instance of ProgressBar
+        The progress bar.
```
larsoner: don't need `Returns` for class

expyfun/visual/_visual.py
```diff
@@ -826,6 +826,76 @@ def __init__(self, ec, colors=('w', 'k')):
...
+
+    This uses two rectangles, one outline, and one solid to show how much
+    progress has been made in the experiment.
+
+    Parameters
+    ----------
+    ec : instance of ExperimentController
+        Parent EC.
+    pos : array-like
+        4-element array-like with X, Y center and width, height.
```
drammock: this could be clearer: `4-element array-like (x, y, width, height) where x and y are coordinates of the box center.`

expyfun/visual/_visual.py
```diff
@@ -826,6 +826,76 @@ def __init__(self, ec, colors=('w', 'k')):
...
+    units : str
+        Units to use. These will apply to all spatial aspects of the drawing.
+        Must be either ''norm'' or ''pix''.
+    colors : list or tuple of matplotlib Colors
+        Colors to fill and outline the bar respectively. Defaults to green and
+        white.
+    text : bool
+        Whether to show the percent done as text below the bar. If ''True'',
```
drammock: shouldn't this be double-backticks, not doubled single quotes?

expyfun/visual/_visual.py
```diff
@@ -826,6 +826,76 @@ def __init__(self, ec, colors=('w', 'k')):
...
+
+    Parameters
+    ----------
+    ec : instance of ExperimentController
+        Parent EC.
+    pos : array-like
+        4-element array-like with X, Y center and width, height.
+    units : str
+        Units to use. These will apply to all spatial aspects of the drawing.
+        Must be either ''norm'' or ''pix''.
```
drammock: I think this should be: `backtick backtick singlequote norm singlequote backtick backtick or backtick backtick singlequote pix singlequote backtick backtick`

expyfun/visual/_visual.py
```diff
@@ -826,6 +826,76 @@ def __init__(self, ec, colors=('w', 'k')):
...
+    Returns
+    -------
+    bar : instance of ProgressBar
+        The progress bar.
+    """
+    def __init__(self, ec, pos, units='norm', colors=('g', 'w')):
+        self._ec = ec
+        if len(colors) != 2:
+            raise ValueError('colors must have length 2')
+        if units != 'norm' and units != 'pix':
```
drammock: @larsoner WDYT about current implementation vs `if units not in ['norm', 'pix']` ?  I definitely tend toward the latter, but it might just be personal style.

expyfun/visual/_visual.py
```diff
@@ -826,6 +826,76 @@ def __init__(self, ec, colors=('w', 'k')):
...
+            raise ValueError('units must be either ''norm'' or ''pix''')
```
drammock: This way of handling quotes doesn't work.  Either use backslashes to escape single quotes within a single-quoted string variable, or use double quotes within single-quoted string variable, or use a double-quoted string variable with single quotes inside it.

expyfun/visual/_visual.py
```diff
@@ -826,6 +826,76 @@ def __init__(self, ec, colors=('w', 'k')):
...
+
+        pos = np.array([float(p) for p in pos])
```
drammock: `np.array(pos, dtype=float)`

expyfun/visual/_visual.py
larsoner: @drammock you can save yourself some pain in this instance by using the triple-backtick code block:
```
``'norm'`` or ``'pix'``
```

expyfun/visual/_visual.py
larsoner: I also prefer the latter

expyfun/visual/_visual.py
maddycapp27: Thanks! And so sorry for my many confusions about convention with the backticks!

expyfun/visual/_visual.py
maddycapp27: Do you want me to change this for Diamond, Rectangle, and FixationDot as well? I agree that it's confusing but assumed it was just me.

larsoner: Sure

expyfun/visual/_visual.py
```diff
@@ -883,9 +881,9 @@ def update_bar(self, percent):
         percent: float
             The percentage of the bar to be filled. Must be between 0 and 1.
         """
-        if percent > 1 or percent < 0:
-            raise ValueError('percent must be a float between 0 and 1')
-        self._pos_bar[2] = percent * self._width
+        if percent > 100 or percent < 0:
+            raise ValueError('percent must be a float between 0 and 100')
+        self._pos_bar[2] = percent * self._width / 100
```
drammock: the divisor should probably be `100.` to be safe...  I'm not positive but it looks possible that both `percent` and `self._width` could be `int`s, which would make this integer division under py2

examples/experiments/progress_bar.py
```diff
@@ -0,0 +1,50 @@
...
+            percent = n * 100 / n_trials
+            pb.update_bar(percent)
+            # display the progress bar with some text
+            ec.screen_text('You\'ve completed {} %. Press any key to proceed.'
+                           ''.format(percent), [0, .1], wrap=False,
+                           font_size=12)
+            pb.draw()
+            ec.flip()
+            # subject uses any key press to proceed
+            ec.wait_one_press()
```
larsoner: This hangs CirceCI:

https://circleci.com/gh/maddycapp27/expyfun/156?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

See how we use `expyfun.building_doc` to prevent this (`git grep building_doc`)

maddycapp27: Ohhhhhhh that makes sense. I knew I had done something wrong but hadn't gotten a chance to figure out what.

examples/experiments/progress_bar.py
```diff
@@ -0,0 +1,53 @@
...
+
+with ExperimentController('name', version='dev', window_size=[500, 500],
+                          full_screen=False, session='foo',
+                          participant='foo') as ec:
+
+    # initialize the progress bar
+    pb = ProgressBar(ec, [0, -.1, 1.5, .1], units='norm')
+
+    ec.screen_prompt('Press the number shown on the screen. Start by pressing'
+                     ' 1.', font_size=12, live_keys=[1])
```
larsoner: [Still hanging](https://circleci.com/gh/maddycapp27/expyfun/157?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

Usually instead of a bunch of conditionals, we do a single one:
```
max_wait = 0.1 if building_doc else np.inf
```
And then pass `max_wait` to all `screen_prompt` (like here) or `wait_one_press` calls.

examples/experiments/progress_bar.py
```diff
@@ -46,8 +47,10 @@
             pb.draw()
             ec.flip()
             # subject uses any key press to proceed
-            if not building_doc:
-                ec.wait_one_press()
+            ec.wait_one_press(max_wait=max_wait)
     ec.screen_text('This example is complete.')
+    screenshot = ec.screenshot()
```
larsoner: It would be a lot cooler in the gallery if you took the screenshot while the visualization was up (e.g., after line 48) on, say, the second trial

maddycapp27: Sorry yes, I (just) realized that the screenshot is for the docs.

examples/experiments/progress_bar.py
```diff
@@ -0,0 +1,57 @@
...
+        number = np.random.randint(1, 5)
+        ec.screen_text(str(number), wrap=False)
+        ec.flip()
+        ec.wait_one_press(live_keys=[number], max_wait=max_wait)
+        ec.flip()
+        ec.wait_secs(wait_dur)
+        # only show progress bar every other trial
+        if n % 2 == 0:
+            # calculate percent done and update the bar object
+            percent = int(n * 100 / n_trials)
```
larsoner: why `int`? `update_bar` should be able to handle `float`

maddycapp27: Oh that's just because I was using it in the text too and didn't want it to say "You completed 33.33333333333333333 %". But there are better ways to do that. I'll fix it first thing tomorrow morning.

larsoner: Yeah it would be better to do it just in the text display, but this is fine for now (I was worried there was some limitation I missed)

larsoner:
The OSX failure was a timeout that I restarted, but those jobs take forever sometimes so I merged. THanks @maddycapp27

maddycapp27:
Ok! Thanks much

PR #322, opened 2017-07-27
MRG: Tracker behavior at floor/ceiling

maddycapp27:
PR to address issues with tracker hitting floors or ceilings. Currently uses ``repeat_limit`` parameter to determine whether to count as ``reversals`` or ``ignore``.

### DIFF
--- a/expyfun/stimuli/_tracker.py
+++ b/expyfun/stimuli/_tracker.py
@@ -122,7 +122,8 @@ class TrackerUD(object):
     """
     def __init__(self, callback, up, down, step_size_up, step_size_down,
                  stop_reversals, stop_trials, start_value, change_indices=None,
-                 change_rule='reversals', x_min=None, x_max=None):
+                 change_rule='reversals', x_min=None, x_max=None,
+                 repeat_limit='reversals'):
         self._callback = _check_callback(callback)
         self._up = up
         self._down = down
@@ -177,7 +178,7 @@ class TrackerUD(object):
         self._n_trials = 0
         self._n_reversals = 0
         self._stopped = False
-        self._repeat_limit = 'reversal'
+        self._repeat_limit = repeat_limit
 
         # Now write the initialization data out
         self._tracker_id = id(self)
@@ -238,6 +239,20 @@ class TrackerUD(object):
                     self._n_reversals += 1
                 if self._direction <= 0:
                     self._direction = 1
+
+        if self._repeat_limit == 'reversals':
+            if self._x[-1] == self._x_min:
+                step_dir = 1
+                self._n_up = 0
+                reversal = True
+                self._n_reversals += 1
+                self._direction = 1
+            if self._x[-1] == self._x_max:
+                step_dir = -1
+                self._n_up = 0
+                reversal = True
+                self._n_reversals += 1
+                self._direction = 1
 
         if reversal:
             self._reversals = np.append(self._reversals, self._n_reversals)
--- a/expyfun/stimuli/_tracker.py
+++ b/expyfun/stimuli/_tracker.py
@@ -522,7 +522,7 @@ class TrackerUD(object):
             return np.nan
         else:
             if any([x in [self._x_min, self._x_max] for x in
-                    self._x[rev_inds]]) and self._repeat_limit=='reversals':
+                    self._x[rev_inds]]) and self._repeat_limit == 'reversals':
                 raise ValueError('Cannot calculate thresholds with reversals '
                                  'at x_min or x_max. Try increasing n_skip.')
             return (np.mean(self._x[rev_inds[0::2]]) +
--- a/expyfun/stimuli/_tracker.py
+++ b/expyfun/stimuli/_tracker.py
@@ -523,8 +523,8 @@ class TrackerUD(object):
         if len(rev_inds) < 1:
             return np.nan
         else:
-            if any([x in [self._x_min, self._x_max]
-                    for x in self._x[rev_inds]]):
+            if any([x in [self._x_min, self._x_max] for x in 
+                    self._x[rev_inds]]) and self._repeat_limit=='reversals':
                 raise ValueError('Cannot calculate thresholds with reversals '
                                  'at x_min or x_max. Try increasing n_skip.')
             return (np.mean(self._x[rev_inds[0::2]]) +
--- a/expyfun/stimuli/_tracker.py
+++ b/expyfun/stimuli/_tracker.py
@@ -272,8 +272,18 @@ class TrackerUD(object):
 
         if not self._stopped:
             if self._x_min is not None:
+                if self._x[-1] < self._x_min:
+                    self._limit_count += 1
+                    if self._repeat_limit == 'reversals':
+                        reversal = True
+                        self._n_reversals += 1  
                 self._x[-1] = max(self._x_min, self._x[-1])
             if self._x_max is not None:
+                if self._x[-1] > self._x_max:
+                    self._limit_count += 1
+                    if self._repeat_limit == 'reversals':
+                        reversal = True
+                        self._n_reversals += 1  
                 self._x[-1] = min(self._x_max, self._x[-1])
 
             self._x_current = self._x[-1]
@@ -314,7 +324,7 @@ class TrackerUD(object):
         else:
             self._n_stop = False
         if self._n_stop and self._limit_count > 0:
-            warnings.warn('Tracker {} hit x_min or x_max {} times.'
+            warnings.warn('Tracker {} exceeded x_min or x_max bounds {} times.'
                           ''.format(self._tracker_id, self._limit_count))
         return self._n_stop
 
--- a/expyfun/stimuli/_tracker.py
+++ b/expyfun/stimuli/_tracker.py
@@ -95,9 +95,9 @@ class TrackerUD(object):
     x_max : float
         The maximum value that the tracker level (``x``) is allowed to take.
     repeat_limit : str
-        How to treat trials that hit either x_min or x_max. `reversals` will
+        How to treat trials that hit either x_min or x_max. ``reversals`` will
         count these trials as reversals and increase or decrease the level
-        respectively. `ignore` does not automatically count these trials as
+        respectively. ``ignore`` does not automatically count these trials as
         reversals and will not change the level of the tracker until a
         correct or incorrect response respectively.
 
--- a/expyfun/stimuli/_tracker.py
+++ b/expyfun/stimuli/_tracker.py
@@ -293,6 +293,24 @@ class TrackerUD(object):
                     reversals=[int(s) for s in self._reversals],
                     x=[float(s) for s in self._x])))
 
+    def check_valid(self, n_reversals):
+        """See if the last reversals contain x_min/x_max reversals.
+
+        Parameters
+        ----------
+        n_reversals : int
+            Number of reversals (starting from the end to check)
+
+        Returns
+        -------
+        valid : bool
+            True if none of the reversals are at x_min or x_max and False
+            otherwise.
+        """
+        self._valid = np.invert(any([x == self._x_min or x == self._x_max
+                                     for x in self._x[-n_reversals:]]))
+        return self._valid
+
     def _stop_here(self):
         if self._n_reversals > self._stop_reversals:
             self._n_stop = True
@@ -300,8 +318,8 @@ class TrackerUD(object):
             self._n_stop = True
         else:
             self._n_stop = False
-        if self._n_stop:
-            warnings.warn('Tracker {} hit a x_min or x_max {} times'
+        if self._n_stop and self._limit_count > 0:
+            warnings.warn('Tracker {} hit a x_min or x_max {} times.'
                           ''.format(self._tracker_id, self._limit_count))
         return self._n_stop
 
--- a/expyfun/stimuli/_tracker.py
+++ b/expyfun/stimuli/_tracker.py
@@ -94,6 +94,12 @@ class TrackerUD(object):
         The minimum value that the tracker level (``x``) is allowed to take.
     x_max : float
         The maximum value that the tracker level (``x``) is allowed to take.
+    repeat_limit : str
+        How to treat trials that hit either x_min or x_max. `reversals` will
+        count these trials as reversals and increase or decrease the level
+        respectively. `ignore` does not automatically count these trials as
+        reversals and will not change the level of the tracker until a
+        correct or incorrect response respectively.
 
     Returns
     -------
@@ -299,7 +305,7 @@ class TrackerUD(object):
         Parameters
         ----------
         n_reversals : int
-            Number of reversals (starting from the end to check)
+            Number of reversals (starting from the end to check).
 
         Returns
         -------
@@ -307,8 +313,10 @@ class TrackerUD(object):
             True if none of the reversals are at x_min or x_max and False
             otherwise.
         """
-        self._valid = np.invert(any([x == self._x_min or x == self._x_max
-                                     for x in self._x[-n_reversals:]]))
+        
+        idx = np.where([r != 0 for r in self._reversals])[0]
+        self._valid = not any(x in (self._x_min, self._x_max) 
+                              for x in self._x[idx[-n_reversals:]])
         return self._valid
 
     def _stop_here(self):
--- a/expyfun/stimuli/_tracker.py
+++ b/expyfun/stimuli/_tracker.py
@@ -96,10 +96,8 @@ class TrackerUD(object):
         The maximum value that the tracker level (``x``) is allowed to take.
     repeat_limit : str
         How to treat trials that hit either x_min or x_max. ``reversals`` will
-        count these trials as reversals and increase or decrease the level
-        respectively. ``ignore`` does not automatically count these trials as
-        reversals and will not change the level of the tracker until a
-        correct or incorrect response respectively.
+        count these trials as reversals while staying at the same level.
+        ``ignore`` does not automatically count these trials as reversals.
 
     Returns
     -------
@@ -248,11 +246,11 @@ class TrackerUD(object):
                 if self._direction <= 0:
                     self._direction = 1
 
-        if self._repeat_limit == 'reversals':
-            if float(self._x[-1]) in [self._x_min, self._x_max]:
+        if float(self._x[-1]) in [self._x_min, self._x_max]:
+            self._limit_count += 1
+            if self._repeat_limit == 'reversals':
                 reversal = True
                 self._n_reversals += 1
-                self._limit_count += 1
 
         if reversal:
             self._reversals = np.append(self._reversals, self._n_reversals)
@@ -523,7 +521,7 @@ class TrackerUD(object):
         if len(rev_inds) < 1:
             return np.nan
         else:
-            if any([x in [self._x_min, self._x_max] for x in 
+            if any([x in [self._x_min, self._x_max] for x in
                     self._x[rev_inds]]) and self._repeat_limit=='reversals':
                 raise ValueError('Cannot calculate thresholds with reversals '
                                  'at x_min or x_max. Try increasing n_skip.')
--- a/expyfun/stimuli/_tracker.py
+++ b/expyfun/stimuli/_tracker.py
@@ -261,16 +261,16 @@ class TrackerUD(object):
         if self._x_min is not None:
             if self._x[-1] < self._x_min:
                 self._limit_count += 1
+                bad = True
                 if self._repeat_limit == 'reversals':
                     reversal = True
-                    bad = True
                     self._n_reversals += 1
         if self._x_max is not None:
             if self._x[-1] > self._x_max:
                 self._limit_count += 1
+                bad = True
                 if self._repeat_limit == 'reversals':
                     reversal = True
-                    bad = True
                     self._n_reversals += 1
 
         if reversal:
@@ -303,7 +303,7 @@ class TrackerUD(object):
                     x=[float(s) for s in self._x])))
 
     def check_valid(self, n_reversals):
-        """See if the last reversals contain x_min/x_max reversals.
+        """If last reversals contain reversals exceeding x_min or x_max.
 
         Parameters
         ----------
@@ -317,8 +317,8 @@ class TrackerUD(object):
             otherwise.
         """
         idx = np.where([r != 0 for r in self._reversals])[0]
-        self._valid = not any(float(x) in (self._x_min, self._x_max)
-                              for x in self._x[idx[-n_reversals:]])
+        self._valid = not any([x in self._bad_reversals
+                               for x in self._reversals[idx[-n_reversals:]]])
         return self._valid
 
     def _stop_here(self):
--- a/expyfun/stimuli/_tracker.py
+++ b/expyfun/stimuli/_tracker.py
@@ -184,6 +184,7 @@ class TrackerUD(object):
         self._n_reversals = 0
         self._stopped = False
         self._repeat_limit = repeat_limit
+        self._bad_reversals = np.asarray([], dtype=int)
         self._limit_count = 0
 
         # Now write the initialization data out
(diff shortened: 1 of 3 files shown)


expyfun/stimuli/_tracker.py
```diff
@@ -239,6 +240,20 @@ def respond(self, correct):
...
+                reversal = True
+                self._n_reversals += 1
+                self._direction = 1
+            if self._x[-1] == self._x_max:
+                step_dir = -1
+                self._n_up = 0
+                reversal = True
+                self._n_reversals += 1
+                self._direction = 1
+
```
maddycapp27: @Eric89GXL @drammock @rkmaddox 

There should probably be some warning when the data is doing this. But I can think of a couple of approaches...
1. Be obnoxious and warn user _every_ time they hit a floor/ceiling.
2. Warn the user that a floor/ceiling was hit x times (no matter how big or small x is) when a tracker stops.
3. Warn the user that "enough" floor/ceiling hits occurred when a tracker stops--then again, how does one decide what "enough" is?
I don't think that there are scenarios in which I'd want to abort an experiment because of floor/ceiling hits.

Eric89GXL: I'd say warn when the tracker stops.

Also it might only matter if the last `N` reversals were at floor/ceiling (e.g., usually those at the last step size). So it would be good to expose some property that the experimenter can check like `tr.check_valid(5)` would check to see if the last 5 reversals were valid (non-ceiling, non-floor). WDYT?

maddycapp27: hmm, I like that idea. I feel like it should be an either or (i.e. warn or property), so for ``check_valid`` I'd want it to say either ``True`` for "yay all good" or some ``int`` to tell how many reversals were ceiling/floor which would replace the functionality of the warning.

Eric89GXL: They serve different purposes. A warning notifies of a potential problem (a property can't do this), and the property allows for investigating whether the potential problem is a real problem (a warning can't do this)

maddycapp27: Ah true. Ok!

expyfun/stimuli/_tracker.py
```diff
@@ -274,13 +293,34 @@ def respond(self, correct):
...
+        n_reversals : int
+            Number of reversals (starting from the end to check)
+
+        Returns
+        -------
+        valid : bool
+            True if none of the reversals are at x_min or x_max and False
+            otherwise.
+        """
+        self._valid = np.invert(any([x == self._x_min or x == self._x_max
```
Eric89GXL: Go full NumPy:
```
~((x == self._x_min) | (x == self._x_max)).any()
```
or go full Python and use the generator expression
```
not any(x in (self._x_min, self._x_max) for x in self._x[-n_reversals:]
```

maddycapp27:
Looks like Travis fail is for something else. 

@Eric89GXL @drammock @rkmaddox 
Ready for review. I'll be away next week but will try to work on this when I am able. Thanks!

expyfun/stimuli/tests/test_tracker.py
```diff
@@ -72,8 +73,8 @@ def test_tracker_ud():
     tr = TrackerUD(None, 3, 1, [1, 0.5], [1, 0.5], 10, np.inf, 1,
                    change_indices=[2])
     tr.respond(True)
-    tr = TrackerUD(None, 3, 1, [1, 0.5], [1, 0.5], np.inf, 10, 1,
-                   change_indices=[2], change_rule='trials')
+    tr = TrackerUD(None, 1, 1, [3, 1.5], [1, 0.5], np.inf, 10, 1,
+                   change_indices=[2], change_rule='trials', x_min=0, x_max=3)
```
Eric89GXL: I suspect these tests do not adequately cover the given functionality. It would be good to construct a track (e.g., 5 or 10 trials?), and check that a warning is emitted, that `check_valid` gives True / False depending on how many reversals are used, and that these behaviors depend on `repeat_limit`.

I know this might seem silly/overkill but such more thorough tests often uncover otherwise hidden bugs.

maddycapp27: Ooops, it doesn't seem silly at all. I started to do something along those lines and got distracted.

expyfun/stimuli/tests/test_tracker.py
```diff
@@ -72,11 +73,13 @@ def test_tracker_ud():
...
-    tr = TrackerUD(None, 3, 1, [1, 0.5], [1, 0.5], np.inf, 10, 1,
-                   change_indices=[2], change_rule='trials')
-    for t in np.arange(2):  # run long enough to encounter change_indices
-        tr.respond(True)
-        tr.respond(False)
+    tr = TrackerUD(None, 1, 1, [3, 1.5], [1, 0.5], np.inf, 5, 1,
+                   change_indices=[2], change_rule='trials', x_min=0, x_max=3)
+    responses = [True, False, True, True, False, True]
+    for r in responses:  # run long enough to encounter change_indices
+        tr.respond(r)
```
Eric89GXL: We don't want un-caught warnings in tests, can you use `with warnings.catch_warnings(record=True) as w:` context manager and ensure e.g. `assert_equal(len(w), 1)` and `assert_true('invalid' in str(w[0].message))`?

maddycapp27: Thanks! I wasn't sure how to search for that.

Eric89GXL:
Let me know when I shoudl look again

maddycapp27:
@Eric89GXL Looks fine on my end, feel free to review whenever is convenient for you!

expyfun/stimuli/_tracker.py
```diff
@@ -93,6 +94,12 @@ class TrackerUD(object):
         The minimum value that the tracker level (``x``) is allowed to take.
     x_max : float
         The maximum value that the tracker level (``x``) is allowed to take.
+    repeat_limit : str
+        How to treat trials that hit either x_min or x_max. `reversals` will
```
Eric89GXL: double-backticks to render as code (single-backticks are for links)

expyfun/stimuli/_tracker.py
```diff
@@ -239,6 +248,22 @@ def respond(self, correct):
                 if self._direction <= 0:
                     self._direction = 1
 
+        if self._repeat_limit == 'reversals':
+            if self._x[-1] == self._x_min:
```
Eric89GXL: This conditional seems weird. Let's say we have someone who just did a trial at the floor (dashed line indicating the next level to which they cannot ever go, i.e. `x_min - 1`):
```
\
 \
-----
```
If they get it right, we'd want it to do this:
```
\  
 \/
-----
```
But it seems like in this case no special treatment is required, because it should already be counted as a reversal by the existing logic. And this seems perfectly fine from an experimental validity standpoint.

In case they get it wrong, however, the track should stay at the same level (not get harder):
```
\
 \_
-----
```
But I don't think that's what this does?

rkmaddox: I think there's a couple things flipped here. For an adaptive track, down is harder, so it goes down on correct answers and up on incorrect. So at the minimum, if you get it wrong, it goes up (regardless of ``repeat_limit``.). If you get it right, it will count a reversal and change the direction to up.

Eric89GXL: > If you get it right, it will count a reversal and change the direction to up.

It shouldn't change the direction and get easier though -- it should stay at the bottom.

Eric89GXL: (and count as a reversal despite staying flat)

maddycapp27: @rkmaddox and I were discussing this a few weeks ago and I think our consensus was that having a reversal without a change in direction was a little odd mathematically. @rkmaddox is that still your feeling on the issue?

rkmaddox: Right, yeah that was the issue. If you count a reversal but don't change direction then it creates weird situations where you can be moving down, and then eventually be moving up after an even number of reversals, which shouldn't be possible.

Without the limit, the subject wills stay at the limit until they eventually get a wrong answer and a reversal, moving up. The limit essentially fast-forwards to that point, skipping a bunch of useless correct answers.

Eric89GXL: >  If you count a reversal but don't change direction then it creates weird situations where you can be moving down, and then eventually be moving up after an even number of reversals, which shouldn't be possible.

Sure, but you're essentially at an invalid part of the track, so you shouldn't include any such "static" reversals in your calculations. So I don't see this being a problem mathematically.

> Without the limit, the subject wills stay at the limit until they eventually get a wrong answer and a reversal, moving up. The limit essentially fast-forwards to that point, skipping a bunch of useless correct answers.

But those correct answers aren't useless -- they tell you something about what the subject did, namely got stuck, which you wouldn't see from looking at the immediate-bounce track.

I'm also thinking about this more from the other end, where the task is too difficult. With the current scheme you'd have this:
```
--------
/\/\/\/\
```
But with what I'm proposing you'd see:
```
--------
/‾‾‾‾‾‾
```
The latter seems like what I'd expect the tracker to do when someone keeps getting it wrong.

Eric89GXL: ... and in both cases, you'd record 7 reversals, and not want to use any of them in a subsequent calculation.

Let's take an example where a subject starts out bad, but eventually figures it out. In the current scheme, the full run could look like this (12 reversals):

```
-------------------
  /\/\/\/\
 /        \  /\
/          \/  \/\/
```
My proposal would be that it should look like this:
```
-------------------
  /‾‾‾‾‾‾\
 /        \  /\
/          \/  \/\/
```
I think in either case you should only use the last 5 reversals in a calculation (and get an error if you use the built-in threshold calculators and include these bound-limited points).

Maybe this gets more complicated when it's not 1-up, 1-down, I'm not sure...

rkmaddox: > Maybe this gets more complicated when it's not 1-up, 1-down, I'm not sure...

It almost certainly would... 

Hitting the limits more than just a quick brush is going to hose any threshold estimate you might do, so maybe it isn't worth all this thought.

@maddycapp27 Please do test it with a 2-down 1-up, or even a 3-down, 2-up track to make sure it doesn't do anything crazy.

Eric89GXL:
> Hitting the limits more than just a quick brush is going to hose any threshold estimate you might do, so maybe it isn't worth all this thought.

The most critical thing for this mode is that, if you are at a limit, and the tracker decides that you should in principle exceed that limit (but in practice of course you can't), it should count as a reversal. How about we have two modes: `'reversal-opposite'` (current behavior; i.e., make a reversal in the opposite of the response-indicated direction) and `'reversal-remain'` (my proposal; i.e., stay at the same level instead of going in the opposite direction).

Either way, it's critical that 1) the thresholding function throw an error if you include any of these points in your computation (because it's going to be invalid), and 2) that there be some way of telling quickly (via function) if any of these points are in the last `N` reversals.

maddycapp27:
I've run a bunch of tracks to see if things are even working right as they are. Well...

Here's a 1 up 1 down case which definitely has some bizarreness. The first reversal is a true correct -> incorrect with the remain trials being all incorrect. This does not make sense to me because we have some reversals that aren't from a change in correctness but do have a change in direction and some changes in direction that don't result in reversals.

1up_1down

Here's a similar 1 up 1 down track that has a reversal at trial 5 that corresponds to a change in correctness without a change in direction (this happens every time there is an odd number of incorrects at a ceiling followed by a correct response).

lagging_reversal

Then we have this 2 down 1 up one which has been bumped down at trial 5 from the subject being incorrect at a ceiling and then has another step down from getting two correct answers. This seems wrong too.
 
2down_1up

Similar things going on in the 3 down 1 up case.

3down_1up

This raises the following questions:

In the /\/\/\ case, do we want all of the changes in direction to be reversals-- /*\*/*\*/*\ or only the ceiling ones-- /*\/*\/*\? The former seems more reasonable to me.

How do we deal with the post-odd ceiling hit reversal as in the second case?

What should we do about extra steps down in the N down 1 up cases?

Is it worth counting the floor/ceiling behavior as reversals anymore or are the warnings sufficient?

Eric89GXL:
> This does not make sense to me because we have some reversals that aren't from a change in correctness but do have a change in direction and some changes in direction that don't result in reversals.

This wouldn't happen if we adopted my proposal of staying flat instead of bouncing, right?

maddycapp27:
> This wouldn't happen if we adopted my proposal of staying flat instead of bouncing, right?

I believe so. I will try it out--particularly with the N down 1 up cases to make sure there aren't similar confounding cases.

Eric89GXL:
Okay, but keep in mind that such stay-flat-because-we-can't-exceed-the-limit-but-want-to moves should count as reversals in terms of counting toward the limit / end of the experiment, but we should protect against ever using them to compute a threshold. Make sense?

maddycapp27:
Yup, I'll look at tr.threshold and throw in an error if repeat_limit is reversals.

maddycapp27:
Sorry to keep you in suspense! The no-bounce option seems to be working pretty well.

Eric89GXL:
Have you convinced @rkmaddox? :)

rkmaddox:
I have been convinced through attrition.

rkmaddox:
(Currently trying to figure out why my Ubuntu "upgrade" killed both my CUDA and my MKL)

drammock:
@rkmaddox cue sad trombone...

maddycapp27:
@Eric89GXL @drammock @rkmaddox 
Looks like all the tests are passing now. Let me know if the current functionality is OK.

expyfun/stimuli/_tracker.py
```diff
@@ -93,6 +94,10 @@ class TrackerUD(object):
         The minimum value that the tracker level (``x``) is allowed to take.
     x_max : float
         The maximum value that the tracker level (``x``) is allowed to take.
+    repeat_limit : str
+        How to treat trials that hit either x_min or x_max. ``reversals`` will
+        count these trials as reversals while staying at the same level.
+        ``ignore`` does not automatically count these trials as reversals.
```
larsoner: I think it's less about "How to treat trials that *hit* either x_min or x_max" and more about "How to treat trials that, *based on the responses, should exceed* either x_min or x_max". In other words, touching is no problem / not interesting. What needs to be decided is how to deal with what happens when a user response indicates that they should go in a direction that they cannot. Agreed?

Is this how you have it implemented in the code, namely that you only need to do something special (stay flat, and optionall count as a reversal depending on this ``kwarg``) if the tracker would otherwise try to exceed a limit?

expyfun/stimuli/_tracker.py
```diff
@@ -274,13 +287,35 @@ def respond(self, correct):
                     reversals=[int(s) for s in self._reversals],
                     x=[float(s) for s in self._x])))
 
+    def check_valid(self, n_reversals):
+        """See if the last reversals contain x_min/x_max reversals.
```
larsoner: "reversals" at either extreme should be fine if they actually correspond to responses that cause a reverse in direction. So I don't get this wording

expyfun/stimuli/_tracker.py
maddycapp27: For some reason I thought we didn't want the subject to touch a ceiling or floor at all (probably based on the way I was using it). But what you're saying makes sense. I will make adjustments.

maddycapp27:
@larsoner (?) 
I think I've addressed the issue. I updated the behavior, warning, error, and check_valid function. Anything I missed?

larsoner:
Yes @drammock made me change my username. I'll look tonight or tomorrow.

maddycapp27:
Looks like one of my tests is failing anyway, tomorrow's plenty early for you to look.

drammock:
I was wondering when I'd get blamed for the name change.  Frankly I'm surprised it didn't happen sooner on Gitter.

larsoner:
You mean `git blame`d!

expyfun/stimuli/_tracker.py
larsoner: This doc still needs an update

expyfun/stimuli/_tracker.py
```diff
@@ -254,6 +262,31 @@ def respond(self, correct):
             self._x = np.append(self._x, self._x[-1] +
                                 self._current_step_size_up)
 
+        if bound:
+            if self._x_min is not None:
```
larsoner: wait, it can be `None`? I thought it was set to `np.inf` or `-np.inf` in the constructor

expyfun/stimuli/_tracker.py
```diff
@@ -254,6 +262,31 @@ def respond(self, correct):
...
+                if self._x[-1] < self._x_min:
+                    self._limit_count += 1
+                    bad = True
+                    if self._repeat_limit == 'reversals':
+                        reversal = True
+                        self._n_reversals += 1
```
larsoner: you still need to then actually revert the next (currently illegal) step with `self._x[-1] = self._x_min`, no?

expyfun/stimuli/_tracker.py
```diff
@@ -274,13 +307,35 @@ def respond(self, correct):
...
+
+        Returns
+        -------
+        valid : bool
+            True if none of the reversals are at x_min or x_max and False
+            otherwise.
+        """
+        idx = np.where([r != 0 for r in self._reversals])[0]
+        self._valid = not any([x in self._bad_reversals
+                               for x in self._reversals[idx[-n_reversals:]]])
```
larsoner: What if we make `self._bad_reversals` a boolean array of the same length as `self._reversals` (setting True when there is a bad one), then you can just do:
```
self._valid = not self._bad_reversals[self._reversals != 0][-n_reversals:].any()
```
Is it clearer for you, too?

expyfun/stimuli/tests/test_tracker.py
```diff
@@ -72,11 +74,26 @@ def test_tracker_ud():
...
+        assert_raises(ValueError, tr.threshold)
+    # run tests with ignore too--should generate warnings, but no error
+    with warnings.catch_warnings(record=True) as w:
+        tr = TrackerUD(None, 1, 1, 0.75, 0.25, np.inf, 7, 1,
+                       x_min=0, x_max=2, repeat_limit='ignore')
+        responses = [True, False, False, False, True, True, True, False]
+        for r in responses:  # run long enough to encounter change_indices
+            tr.respond(r)
+        assert_equal(len(w), 1)
+        tr.threshold(n_skip=0)
```
larsoner: if this does not make a warning, unindent it, or leave it indented (as is) but add an `assert_equal(len(w), 1)` afterward, i.e. ensure no new warning is emitted

expyfun/stimuli/tests/test_tracker.py
```diff
@@ -72,11 +74,26 @@ def test_tracker_ud():
...
+    with warnings.catch_warnings(record=True) as w:
+        tr = TrackerUD(None, 1, 1, 0.75, 0.25, np.inf, 7, 1,
+                       x_min=0, x_max=2)
+        responses = [True, False, False, False, True, True, True, False]
+        for r in responses:  # run long enough to encounter change_indices
+            tr.respond(r)
+        assert(tr.check_valid(1))  # make sure checking validity is good
+        assert(not tr.check_valid(3))
+        assert_equal(len(w), 1)
+        assert_raises(ValueError, tr.threshold)
```
larsoner: It's good to have this check, but more complete to have something a check that `tr.threshold(-2)` works (or whatever you need to do to use the last `N` reversals that are valid) and that the value is correct. (If there are no valid reversals at the end, modify the test so that there are a few valid ones at the end that you can use.)

expyfun/stimuli/_tracker.py
```diff
@@ -482,6 +541,11 @@ def threshold(self, n_skip=2):
         if len(rev_inds) < 1:
             return np.nan
         else:
+            if any([x in self._bad_reversals for x in
+                    rev_inds]) and self._repeat_limit == 'reversals':
```
larsoner: vectorize
```
self._bad_reversals[rev_inds].any()
```
although this will change if you adopt my other suggestion

expyfun/stimuli/_tracker.py
maddycapp27: That happens later

expyfun/stimuli/_tracker.py
maddycapp27: Yea, that actually makes a lot of things cleaner

expyfun/stimuli/_tracker.py
```diff
@@ -283,17 +283,15 @@ def respond(self, correct):
...
         # Should we stop here?
         self._stopped = self._stop_here()
 
         if not self._stopped:
-            if self._x_min is not None:
+            if self._x_min is not -np.inf:
                 self._x[-1] = max(self._x_min, self._x[-1])
-            if self._x_max is not None:
+            if self._x_max is not np.inf:
                 self._x[-1] = min(self._x_max, self._x[-1])
```
larsoner: Ahh I see, it's done here. Why not move these lines up to where the `self._x_max` / `self._x_min` checks have already been done? Seems a bit cleaner to do it all at once.

expyfun/stimuli/_tracker.py
larsoner: i.e., put them in these branches because they already exist

maddycapp27: The only reason I hadn't done that is because of the conditional for whether the tracker was stopped. Which on further inspection I'm not really sure why it's there? The next x value obviously won't be used, but it's added to ``self._x`` in lines 256-263 anyway so there's no reason _not_ to constrain it to the ``x_min`` or ``x_max``, right?

larsoner: Agreed

expyfun/stimuli/_tracker.py
```diff
@@ -177,7 +183,9 @@ def __init__(self, callback, up, down, step_size_up, step_size_down,
         self._n_trials = 0
         self._n_reversals = 0
         self._stopped = False
-        self._repeat_limit = 'reversal'
+        self._repeat_limit = repeat_limit
+        self._bad_reversals = np.asarray([], dtype=int)
```
larsoner: dtype=bool

expyfun/stimuli/_tracker.py
```diff
@@ -482,6 +536,10 @@ def threshold(self, n_skip=2):
         if len(rev_inds) < 1:
             return np.nan
         else:
+            if sum(self._bad_reversals[rev_inds]) > 0:
```
larsoner: `if self._bad_reversals[rev_inds].any()`

expyfun/stimuli/tests/test_tracker.py
```diff
@@ -72,11 +74,28 @@ def test_tracker_ud():
...
+
+    with warnings.catch_warnings(record=True) as w:
+        tr = TrackerUD(None, 1, 1, 0.75, 0.75, np.inf, 8, 1,
+                       x_min=0, x_max=2)
+        responses = [True, True, True, False, False, False, False, True, False]
+        for r in responses:  # run long enough to encounter change_indices
+            tr.respond(r)
+        assert_equal(len(w), 1)
+    assert(tr.check_valid(1))  # make sure checking validity is good
+    assert(not tr.check_valid(3))
```
larsoner: technically these should be `assert_true` but we'll probably change over to `py.test` soonish so it's okay

larsoner:
@rkmaddox did you want to look again?

FWIW I just upgraded to Ubuntu 17.04 to see if I could replicate your CUDA/MKL problems but everything seems to work okay over here after reinstalling `PyCUDA` (which needed to be recompiled due to newer CUDA). :(

rkmaddox:
Nah, I'm good.

Huh. Will try compiling `PyCUDA` from source. Didn't have to do that last time. The MKL thing I found very surprising.

larsoner:
If you're using the ones from the Ubuntu repo then it should just work. Weird that it doesn't.

larsoner:
Thanks @maddycapp27

maddycapp27:
Yay! Thanks @larsoner  (also sorry for all the confusion)

larsoner:
@maddycapp27 I have `stop_reversals=10` and (intentionally doing the task terribly) I got this warning:
```
/home/larsoner/Documents/python/larsoner/adaptivetempcoh/expyfun/stimuli/_tracker.py:333: UserWarning: Tracker 140478825957136 exceeded x_min or x_max bounds 11 times.
  ''.format(self._tracker_id, self._limit_count))
```
It should only be 10, right? Or am I missing something? Either the message or the logic is wrong...

maddycapp27:
It's always (as long as I've been using it) done ``n_reversals = stop_reversals + 1`` which I thought was odd but intentional.

larsoner:
This sounds like a bug to me, @rkmaddox is this intentional / what's the rationale?

rkmaddox:
Just kind of how the numbers shake out. A reversal includes the trials that come after it. So if you did a 1-reversal track, you would go down for a while, get one wrong, go up for a while, and the track would stop next time you got one wrong:

```
\
 \
  \  /
   \/
```

The above is a 1 reversal track (a one-reversal track should include one down part and one up part). `n_reversals` will be 2 when the track ends. We could decrement it when the track finishes to allow it to be a little more intuitive?

rkmaddox:
Would also want to decrement `n_trials` at stop (since I think it will be 1 too high as well).

rkmaddox:
Actually, maybe a 1-reversal track should just be down. I am open to changing it.

larsoner:
> Actually, maybe a 1-reversal track should just be down. I am open to changing it.

So a 1-reversal track would mean "a track in which the responses indicated that it's time to change directions once", and `n_reversals == 1`.

Whereas right now a 1-reversal track means "a track in which the step actually changed directions once", and `n_reversals == 2`?

rkmaddox:
Correct.

larsoner:
I think I prefer the former description (one-reversal track is based on what responses indicate to do, and `n_reversals == 1`)? Okay with you @rkmaddox @maddycapp27 ?

rkmaddox:
I also prefer it, after consideration.

maddycapp27:
Makes more intuitive sense that way so I'm happy to work on fixing it.

larsoner:
Sounds like you should be good to get started when you have time

PR #323, opened 2017-08-14
MRG: Fix tracker stopping

maddycapp27:
Issue with tracker stopping at ``stop_trials + 1`` or ``stop_reversals + 1`` rather than at the actual stopping criteria.

### DIFF
--- a/expyfun/stimuli/tests/test_tracker.py
+++ b/expyfun/stimuli/tests/test_tracker.py
@@ -3,7 +3,7 @@ import matplotlib
 matplotlib.use('Agg')  # noqa
 from expyfun.stimuli import TrackerUD, TrackerBinom, TrackerDealer
 from expyfun import ExperimentController
-from nose.tools import assert_raises, assert_equal, assert_true
+from nose.tools import assert_raises, assert_equal
 from expyfun._utils import _hide_window, requires_opengl21
 import warnings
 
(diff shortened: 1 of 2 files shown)


expyfun/stimuli/tests/test_tracker.py
```diff
@@ -27,10 +27,11 @@ def test_tracker_ud():
     tr = TrackerUD(callback, 3, 1, 1, 1, np.inf, 10, 1)
     with ExperimentController('test', **std_kwargs) as ec:
         tr = TrackerUD(ec, 3, 1, 1, 1, np.inf, 10, 1)
-    tr = TrackerUD(None, 3, 1, 1, 1, np.inf, 10, 1)
+    tr = TrackerUD(None, 3, 1, 1, 1, 10, np.inf, 1)
     rand = np.random.RandomState(0)
     while not tr.stopped:
         tr.respond(rand.rand() < tr.x_current)
+    assert_true(tr.n_reversals == tr.stop_reversals)
```
larsoner: better to use `assert_equal` (if it ever fails the error message is nicer)

larsoner:
Thanks for the quick fix @maddycapp27

larsoner:
Just wanted to let y'all know it seems to work nicely, thanks for all of the work! I can get plots like this in a couple dozen lines now:

figure_1

PR #324, opened 2017-08-29
MRG: Add calibration schemes to eyetracker

maddycapp27:
We ran into issues with the HV5 calibration scheme for the eyetracker and wanted to test out some other options (HV9 and H3 so far). 

I'd also like to add HV13 if possible but am not sure what those target coordinates actually are--based on the icon in the eyelink settings window they aren't quite an even grid and I'm still trying to figure those numbers out.

Disclaimer: I'm working on the computer that's actually attached to the eyelink system so all commits will come from rkmaddox for this PR but they are from me. My apologies!

### DIFF
--- a/expyfun/_eyelink_controller.py
+++ b/expyfun/_eyelink_controller.py
@@ -535,7 +535,8 @@ class EyelinkController(object):
         Parameters
         ----------
         ctype : str
-            Type of calibration. Currently only 'HV5' is supported.
+            Type of calibration. Currently 'H3', 'HV5', 'HV9', and 'HV13'
+            are supported.
         horiz : float
             Horizontal distance (left and right, each) to use.
         vert : float
@@ -547,7 +548,7 @@ class EyelinkController(object):
         --------
         EyelinkController.calibrate
         """
-        allowed_types = ['HV5', 'HV9', 'H3', 'HV13']
+        allowed_types = ['H3', 'HV5', 'HV9', 'HV13']
         if ctype not in allowed_types:
             raise ValueError('ctype cannot be "{0}", but must be one of {1}'
                              ''.format(ctype, allowed_types))
--- a/expyfun/_eyelink_controller.py
+++ b/expyfun/_eyelink_controller.py
@@ -547,7 +547,7 @@ class EyelinkController(object):
         --------
         EyelinkController.calibrate
         """
-        allowed_types = ['HV5', 'HV9', 'H3']
+        allowed_types = ['HV5', 'HV9', 'H3', 'HV13']
         if ctype not in allowed_types:
             raise ValueError('ctype cannot be "{0}", but must be one of {1}'
                              ''.format(ctype, allowed_types))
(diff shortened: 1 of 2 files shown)


rkmaddox:
Awesome--are these calibrations indeed working with the EL?

maddycapp27:
@rkmaddox Yup! HV9 seems better than HV5. Not sure what H3 is really for but hey it was easy to throw in.

rkmaddox:
Excellent! How often are things that are *supposed* to be easy *actually* easy?

maddycapp27:
Wow that was also easy. Feeling like I'm tempting fate here...

maddycapp27:
@rkmaddox @drammock @larsoner Looks ok on my end. Ready for review. Thanks!

expyfun/_eyelink_controller.py
```diff
@@ -547,7 +547,7 @@ def custom_calibration(self, ctype='HV5', horiz=2./3., vert=2./3.,
         --------
         EyelinkController.calibrate
         """
-        allowed_types = ['HV5']
+        allowed_types = ['HV5', 'HV9', 'H3', 'HV13']
```
rkmaddox: Order bugs me :-) Let's do H3, HV5, HV9, HV13.

Also, can't comment there, but please change docs to mention supported types.

rkmaddox:
Assuming tests pass and you have tested it on the EL and nose, I am +1 for merge. It would be great if the UW crew had a chance to try it out to make sure it works across systems.

expyfun/_eyelink_controller.py
```diff
@@ -561,7 +562,17 @@ def custom_calibration(self, ctype='HV5', horiz=2./3., vert=2./3.,
...
+            mat = np.array([[0, 0], [1, 0], [-1, 0], [0, 1], [0, -1]])
+        elif ctype == 'HV9':
+            mat = np.array([[0, 0], [1, 0], [-1, 0], [0, 1], [0, -1], [1, 1],
+                            [-1, -1], [1, -1], [-1, 1]])
+        elif ctype == 'H3':
+            mat = np.array([[0, 0], [1, 0], [-1, 0]])
+        elif ctype == 'HV13':
+            mat = np.array([[0, 0], [1, 0], [-1, 0], [0, 1], [0, -1], [1, 1],
+                            [-1, -1], [1, -1], [-1, 1], [.5, .5], [-.5, -.5],
+                            [.5, -.5], [-.5, .5]])
```
larsoner: FYI these `mat` arrays will have different `dtype` (first four `int` last one `float`) because of their inputs which in principle could be dangerous, but it's probably okay :)

larsoner:
Thanks @maddycapp27

larsoner:
> It would be great if the UW crew had a chance to try it out to make sure it works across systems.

Agreed, @drammock if you have time the next time you're messing with the eye-tracker this would be great. In the meantime this preserves the default (and the only previously allowed value) so we should be okay at least with existing experiments.

drammock:
I've put it on my calendar to test these next week.

PR #325, opened 2017-08-30
FIX: Eyelink not able to get position for right eye

maddycapp27:
We have not been able to use the right eye for any tracking until I found what looks to be missing parentheses.

Is this just a fix for us?


maddycapp27:
@rkmaddox @drammock @larsoner 

I'm so confused but having changed this we can use ``el.get_eye_position()`` when tracking the right eye (whereas we couldn't before and spent a decent amount of time trying to figure out why).

rkmaddox:
Yep, that should fix it. The problem is that without the parentheses, `sample.isLeftSample` is a function, not the returned value of the function. A function != 0, and anything that is non-zero evaluates to true. So it always tried to get the left eye position.

larsoner:
Ouch, I think that was my bad from way back when

PR #328, opened 2017-09-01
MRG: Save and report the version

larsoner:
Ready for review/merge from my end.

@maddycapp27 can you look to see if you agree this would be useful?

### DIFF
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -6,12 +6,14 @@
 #
 # License: BSD (3-clause)
 
-import numpy as np
+import json
 import os
 import warnings
 from os import path as op
 from functools import partial
 import traceback as tb
+
+import numpy as np
 
 from ._utils import (get_config, verbose_dec, _check_pyglet_version, wait_secs,
                      running_rms, _sanitize, logger, ZeroClock, date_str,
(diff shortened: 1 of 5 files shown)


maddycapp27:
I like it. Seems like a minor change with a lot of potential to avert confusion.

rkmaddox:
One thing I have also wondered: we write some of the parameters to the tab/log files. Shouldn't we do all of them? `stim_db`, `noise_level`, `targ_rms`, etc.? Otherwise it needs to be noted and recorded elsewhere, which defeats the purpose. These should also of course be logged when changes are made to them as well.

rkmaddox:
Could even do `__file__` (in a `try`/`except`) so that the exact script used to generate the data file is known.

larsoner:
Yeah we should be able to get the calling `__file__` from the stack. I'll work on adding those, setting to WIP in the meantime.

larsoner:
Okay @rkmaddox added logging of everything

@drammock can you look, too, when you get a chance? This seems like an important missing feature

rkmaddox:
LGTM except it still does not say which experimental file was run. Should we add that?

larsoner:
@rkmaddox done

rkmaddox:
Thanks. +1 for merge.

rkmaddox:
Ping @drammock since I don't have merge privileges.

drammock:
I'm happy with this.  tests pass locally and version is correctly reported in the stdout of `simple_experiment.py` (`Using version 2.0.0.dev0 (requested dev)`).  I made a tiny change that I'll PR into this PR momentarily; if @larsoner is happy with that he can merge this one too.

larsoner:
Thanks for the quick reviews!

rkmaddox:
Thanks for the addition. Makes it a much better system, especially when coming back to old data...

PR #329, opened 2017-09-08
MRG: maintain_fix and more custom calibration

maddycapp27:
This adds a function to track fixation over time called ``maintain_fix`` which makes sure you're looking in the correct area at least once per interval.

Additionally, I've added a option to have a truly custom calibration with specified coordinates.

Both still need error checking and tests.

### DIFF
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -1069,6 +1069,10 @@ class ExperimentController(object):
     def check_force_quit(self):
         """Check to see if any force quit keys were pressed."""
         self._response_handler.check_force_quit()
+    
+    def check_interrupts(self):
+        """Check to see if any interrupt keys were pressed."""
+        self._response_handler.check_interrupts()
 
 # ############################## MOUSE METHODS ################################
     def listen_clicks(self):
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -1069,10 +1069,6 @@ class ExperimentController(object):
     def check_force_quit(self):
         """Check to see if any force quit keys were pressed."""
         self._response_handler.check_force_quit()
-    
-    def check_interrupts(self):
-        """Check to see if any interrupt keys were pressed."""
-        self._response_handler.check_interrupts()
 
 # ############################## MOUSE METHODS ################################
     def listen_clicks(self):
--- a/expyfun/_eyelink_controller.py
+++ b/expyfun/_eyelink_controller.py
@@ -620,9 +620,6 @@ class EyelinkController(object):
             if coordinates is not None:
                 raise ValueError('If ctype is not \'custom\' coordinates canno'
                                  't be used to generate calibration pattern.')
-        else:
-            if not isinstance(coordinates, list):
-                raise ValueError('Coordinates must be in a list.')
 
         horiz, vert = float(horiz), float(vert)
         xx = np.array(([0., horiz], [0., vert]))
@@ -646,10 +643,8 @@ class EyelinkController(object):
                             [-1, -1], [1, -1], [-1, 1], [.5, .5], [-.5, -.5],
                             [.5, -.5], [-.5, .5]])
         elif ctype == 'custom':
-            if not isinstance(coordinates, list):
-                raise ValueError('Coordinates must be in a list.')
             mat = np.array(coordinates, float)
-            if mat.shape[-1] != 2:
+            if  mat.ndim != 2 or mat.shape[-1] != 2:
                 raise ValueError('Each coordinate must be a list with length 2'
                                  '.')
         offsets = mat * np.array([h_pix, v_pix])
--- a/expyfun/_eyelink_controller.py
+++ b/expyfun/_eyelink_controller.py
@@ -623,9 +623,7 @@ class EyelinkController(object):
         else:
             if not isinstance(coordinates, list):
                 raise ValueError('Coordinates must be in a list.')
-            if any([not isinstance(c, list) or len(c) != 2
-                    for c in coordinates]):
-                
+
         horiz, vert = float(horiz), float(vert)
         xx = np.array(([0., horiz], [0., vert]))
         h_pix, v_pix = np.diff(self._ec._convert_units(xx, units, 'pix'),
--- a/expyfun/_eyelink_controller.py
+++ b/expyfun/_eyelink_controller.py
@@ -644,7 +644,7 @@ class EyelinkController(object):
                             [.5, -.5], [-.5, .5]])
         elif ctype == 'custom':
             mat = np.array(coordinates, float)
-            if  mat.ndim != 2 or mat.shape[-1] != 2:
+            if mat.ndim != 2 or mat.shape[-1] != 2:
                 raise ValueError('Each coordinate must be a list with length 2'
                                  '.')
         offsets = mat * np.array([h_pix, v_pix])
--- a/expyfun/_eyelink_controller.py
+++ b/expyfun/_eyelink_controller.py
@@ -615,9 +615,17 @@ class EyelinkController(object):
         if ctype not in allowed_types:
             raise ValueError('ctype cannot be "{0}", but must be one of {1}'
                              ''.format(ctype, allowed_types))
-        if ctype != 'custom' and not coordinates == None:
-            raise ValueError('If ctype is \'custom\' coordinates must be '
-                             'specified.')
+        if ctype != 'custom':
+            if not coordinates == None:
+                raise ValueError('If ctype is not \'custom\' coordinates canno'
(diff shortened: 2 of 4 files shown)


expyfun/_eyelink_controller.py
```diff
@@ -615,9 +615,17 @@ def custom_calibration(self, ctype='HV5', horiz=2./3., vert=2./3.,
...
+            if not coordinates == None:
+                raise ValueError('If ctype is not \'custom\' coordinates canno'
+                                 't be used to generate calibration pattern.')
+        else:
+            if isinstance(coordinates, list):
+                raise ValueError('Coordinates must be in a list.')
+            if any([not isinstance(c, list) or len(c) != 2 
+                    for c in coordinates]):
+                raise ValueError('Each coordinate must be a list with length 2'
+                                '.')
```
maddycapp27: Should there also be a stipulation that there should be a minimum number of coordinates? I'm assuming that the eyetracker can't do much with less than 3.

maddycapp27:
@larsoner @drammock @rkmaddox 

This also seems to be an unrelated AppVeyor failure. Not sure what to do about coveralls since I'm getting this error "The file "expyfun/_eyelink_controller.py" isn't available on github. Either it's been removed, or the repo root directory needs to be updated." when I try to open it. Regardless I think I'm ready for review on this one as well.

expyfun/_eyelink_controller.py
```diff
@@ -528,30 +528,105 @@ def wait_for_fix(self, fix_pos, fix_time=0., tol=100., max_wait=np.inf,
...
+        `period` seconds.
+
+        Parameters
+        ----------
+        fix_pos : tuple (length 2)
+            The screen position (in specified units) required.
+        check_duration : float
+            Total amount of time to check.
+        tol : float
+            The tolerance (in pixels) to consider the target hit.
```
larsoner: We usually allow units for these sorts of things. Maybe this should be in the same units as `fix_pos`. And is it a radius (circular area) or w/h (square)? (I should be able to tell from the var name ideally, and if not there, then the description.)

expyfun/_eyelink_controller.py
```diff
@@ -528,30 +528,105 @@ def wait_for_fix(self, fix_pos, fix_time=0., tol=100., max_wait=np.inf,
...
         if ctype not in allowed_types:
             raise ValueError('ctype cannot be "{0}", but must be one of {1}'
                              ''.format(ctype, allowed_types))
+        if ctype != 'custom':
+            if coordinates is not None:
+                raise ValueError('If ctype is not \'custom\' coordinates canno'
+                                 't be used to generate calibration pattern.')
+        else:
+            if isinstance(coordinates, list):
+                raise ValueError('Coordinates must be in a list.')
```
larsoner: if it's a list, you raise an error that it must be a list ... ?

expyfun/_eyelink_controller.py
```diff
@@ -528,30 +528,105 @@ def wait_for_fix(self, fix_pos, fix_time=0., tol=100., max_wait=np.inf,
...
+            if any([not isinstance(c, list) or len(c) != 2
+                    for c in coordinates]):
+                raise ValueError('Each coordinate must be a list with length 2'
+                                 '.')
```
larsoner: The easier check is to do `coordinates = np.array(coordinates, float)` and then check `coordinates.shape`

expyfun/_eyelink_controller.py
maddycapp27: OOPS!

expyfun/_eyelink_controller.py
maddycapp27: ah ok, makes sense

expyfun/_eyelink_controller.py
maddycapp27: I was using ``wait_for_fix()`` as a template which has allows any units for fix_pos but _only_ does tolerance in pixels.

maddycapp27: Do we want to change both ``wait_for_fix()`` and ``maintain_fix()`` to allow for any units?

larsoner: Hmm, that's a good question. Does it seem worth it to you?

maddycapp27: Well, the issue is that the tolerances in pixels make sense anywhere on the screen but if you have fixation dots off to the side and are trying to use degrees as tolerances, things get a little funky. So if we allowed for any units I'd say we'd end up converting whatever tolerances in degrees or norm to pixels at the center and then use those.

I think it's worth it as long as we're transparent about the units not being true on the sides.

larsoner: Good point. We can keep pixels for now, then, and add units later if we need them

larsoner:
Something is wrong with your development workflow, you are pushing commits as @rkmaddox

maddycapp27:
Pushing as @rkmaddox because I'm on the behavioral computer in the lab rather than at my desktop.

larsoner:
It would be good to figure out a better system in the not too distant future, otherwise things like `git blame` don't work correctly.

expyfun/_experiment_controller.py
```diff
@@ -1069,6 +1069,10 @@ def _log_presses(self, pressed):
     def check_force_quit(self):
         """Check to see if any force quit keys were pressed."""
         self._response_handler.check_force_quit()
+    
+    def check_interrupts(self):
+        """Check to see if any interrupt keys were pressed."""
+        self._response_handler.check_interrupts()
```
larsoner: Shouldn't be part of this PR, right?

maddycapp27: Yes, that is my mistake

maddycapp27: (serves me right for trying to multitask)

expyfun/_eyelink_controller.py
```diff
@@ -528,30 +528,102 @@ def wait_for_fix(self, fix_pos, fix_time=0., tol=100., max_wait=np.inf,
...
         if ctype not in allowed_types:
             raise ValueError('ctype cannot be "{0}", but must be one of {1}'
                              ''.format(ctype, allowed_types))
+        if ctype != 'custom':
+            if coordinates is not None:
+                raise ValueError('If ctype is not \'custom\' coordinates canno'
+                                 't be used to generate calibration pattern.')
+        else:
+            if not isinstance(coordinates, list):
+                raise ValueError('Coordinates must be in a list.')
```
larsoner: you don't need this check, `np.array` is good enough (you should really allow list, tuple, ndarray, etc.)

larsoner: (also this is redundant with the check that is done later)

expyfun/_eyelink_controller.py
```diff
@@ -573,6 +645,13 @@ def custom_calibration(self, ctype='HV5', horiz=2./3., vert=2./3.,
             mat = np.array([[0, 0], [1, 0], [-1, 0], [0, 1], [0, -1], [1, 1],
                             [-1, -1], [1, -1], [-1, 1], [.5, .5], [-.5, -.5],
                             [.5, -.5], [-.5, .5]])
+        elif ctype == 'custom':
+            if not isinstance(coordinates, list):
+                raise ValueError('Coordinates must be in a list.')
```
larsoner: you don't need this check

expyfun/_eyelink_controller.py
```diff
@@ -573,6 +645,13 @@ def custom_calibration(self, ctype='HV5', horiz=2./3., vert=2./3.,
...
+            mat = np.array(coordinates, float)
+            if mat.shape[-1] != 2:
```
larsoner: `if mat.ndim != 2 or mat.shape[-1] != 2`

drammock:
I don't have time for an in-depth review right now, but running `make test` locally, it fails flake:

```
Running flake8
expyfun/_eyelink_controller.py:647:15: E271 multiple spaces after keyword
expyfun/_input_controllers.py:170:51: W291 trailing whitespace
```

I also get some warnings (these may not be new):
```
Test EC methods. ... /home/drmccloy/Builds/expyfun/expyfun/_experiment_controller.py:1380: RuntimeWarning: invalid value encountered in multiply
  samples *= self._stim_scaler
/home/drmccloy/Builds/expyfun/expyfun/_sound_controllers.py:151: RuntimeWarning: invalid value encountered in less
  data[data < -1] = -1
/home/drmccloy/Builds/expyfun/expyfun/_sound_controllers.py:152: RuntimeWarning: invalid value encountered in greater
  data[data > 1] = 1
```

larsoner:
Yeah those warnings are old

larsoner:
I'll go ahead and merge because it's just OSX we're waiting for, thanks @maddycapp27

PR #331, opened 2017-09-08
MRG: advance from screen_prompt with click

maddycapp27:
This just adds a few extra lines in ``ec.screen_prompt`` to allow for subjects to click to advance rather than press a key

### DIFF
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -442,9 +442,9 @@ class ExperimentController(object):
         return scr_txt
 
     def screen_prompt(self, text, max_wait=np.inf, min_wait=0, live_keys=None,
-                      timestamp=False, clear_after=True, pos=[0, 0],
-                      color='white', font_name='Arial', font_size=24,
-                      wrap=True, units='norm', attr=True):
+                      click=False, timestamp=False, clear_after=True,
+                      pos=[0, 0], color='white', font_name='Arial',
+                      font_size=24, wrap=True, units='norm', attr=True):
         """Display text and (optionally) wait for user continuation
 
         Parameters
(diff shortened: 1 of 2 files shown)


expyfun/_experiment_controller.py
```diff
@@ -442,9 +442,9 @@ def screen_text(self, text, pos=[0, 0], color='white', font_name='Arial',
         return scr_txt
 
     def screen_prompt(self, text, max_wait=np.inf, min_wait=0, live_keys=None,
-                      timestamp=False, clear_after=True, pos=[0, 0],
-                      color='white', font_name='Arial', font_size=24,
-                      wrap=True, units='norm', attr=True):
+                      click=False, timestamp=False, clear_after=True,
```
larsoner: new params should go at the end to avoid backward compat break with `args` usage

maddycapp27:
@larsoner @drammock @rkmaddox 

Looks like AppVeyor isn't failing for any particular reason. Review?

expyfun/_experiment_controller.py
```diff
@@ -507,8 +509,12 @@ def screen_prompt(self, text, max_wait=np.inf, min_wait=0, live_keys=None,
...
                              attr=attr)
             self.flip()
-            out = self.wait_one_press(max_wait, min_wait, live_keys,
-                                      timestamp)
+            if click:
+                out = self.wait_one_click(max_wait, min_wait, live_keys,
+                                          timestamp)
+            else:
+                out = self.wait_one_press(max_wait, min_wait, live_keys,
+                                          timestamp)
```
larsoner: If you want to be fancy you could do:
```
fun = self.wait_one_click if click else self.wait_one_press
fun(max_wait, min_wait, live_keys, timestamp)
```

larsoner:
Can you add a simple test with `click=True` (probably with `max_wait=1e-3` or something)?

larsoner:
I strongly recommend using a PEP8-checking IDE of some sort

https://travis-ci.org/LABSN/expyfun/jobs/278279182

larsoner:
Thanks @maddycapp27

PR #332, opened 2017-09-08
FIX: Tracker issues

maddycapp27:
Currently just here to fix the lack of ValueErrors for ``up`` and ``down`` in ``TrackerBinom``

### DIFF
--- a/expyfun/stimuli/_tracker.py
+++ b/expyfun/stimuli/_tracker.py
@@ -209,7 +209,8 @@ class TrackerUD(object):
             change_indices=[int(s) for s in self._change_indices],
             change_rule=self._change_rule,
             x_min=self._x_min,
-            x_max=self._x_max)))
+            x_max=self._x_max,
+            repeat_limit=self._repeat_limit)))
 
     def respond(self, correct):
         """Update the tracker based on the last response.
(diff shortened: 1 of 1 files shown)


expyfun/stimuli/_tracker.py
```diff
@@ -130,7 +130,11 @@ def __init__(self, callback, up, down, step_size_up, step_size_down,
                  change_rule='reversals', x_min=None, x_max=None,
                  repeat_limit='reversals'):
         self._callback = _check_callback(callback)
+        if type(up) != int:
```
larsoner: the nicer check would be `if not isinstance(up, int)`. But going one level further, based on a (long) SciPy discussion the consensus is this is the best int-like check:
```
up = operator.index(up)
```
It automatically converts to `int` things like `np.array(12, np.int32)` and throws a reasonable error for non-int-like

maddycapp27: Hmm, so tradeoff is between using ``isinstance`` and having a more specific error message or ``operator.index`` and getting a conversion for int-likes? I think I prefer the former since the tracker code is so lengthy to sort through.

larsoner: > I think I prefer the former since the tracker code is so lengthy to sort through.

`up = operator.index(up)` is one line vs the two for `isinstance` check, and works in more cases, but if you really like `isinstance` we can keep it here

maddycapp27:
@larsoner @drammock 

I found another omission in the tracker code but it should be fixed and ready to merge pending your review

larsoner:
Changes look good and circle failure is unrelated, thanks

Issue #327, opened 2017-09-01
BUG: Can't reconstruct with `reconstruct_tracker`


larsoner:
I have a paradigm with `TrackerDealer` having two `TrackerUD` instances, and when I try to use `reconstruct_tracker` I get this error:
```
  File "/home/larsoner/Documents/python/larsoner/adaptivetempcoh/behavioral.py", line 18, in <module>
    trackers = reconstruct_tracker(fname_tab)
  File "expyfun/io/_parse.py", line 154, in reconstruct_tracker
    tr[-1].respond(r)
  File "expyfun/stimuli/_tracker.py", line 219, in respond
    raise RuntimeError('Tracker is stopped.')
RuntimeError: Tracker is stopped.
```
@maddycapp27 is this use case intended to be supported by `reconstruct_tracker`, and if so, this is a bug, right?

maddycapp27:
I have a few thoughts. Since we recently changed trackers to stop at n trials (or reversals) rather than n + 1 trials (or reversals) I could see this being an issue for data collected before the fix and reconstructed after the fix. My other thought is that if you're using a ``TrackerDealer`` you may want to run ``reconstruct_dealer`` instead -- there's no reason that ``reconstruct_tracker`` wouldn't work though (particularly since ``reconstruct_dealer`` calls ``reconstruct_tracker``).

rkmaddox:
Ahhh, yeah the reconstruction code will need to be updated to reflect the additional arguments.

rkmaddox:
And the new arguments will need to be written to the `init` parts of the expyfun log.

larsoner:
> I could see this being an issue for data collected before the fix and reconstructed after the fix.

I'm pretty sure the code was run after the fix.

> My other thought is that if you're using a TrackerDealer you may want to run reconstruct_dealer instead 

Same error.

> the reconstruction code will need to be updated to reflect the additional arguments.

Which arguments? Do you mean the new arguments / new stopping behavior? If so I'm not sure why it wasn't caught by some test.

larsoner:
> And the new arguments will need to be written to the init parts of the expyfun log.

I just used the default for `repeat_limit`, though

maddycapp27:
Hmm if you ran the original trackers after we changing the stopping criteria for the trackers then the length of the responses wouldn't be the source of the issue...

The new arguments shouldn't be an issue. It just makes a dict from the log file to fill in kwargs.

rkmaddox:
[...]

PR #333, opened 2017-09-12
FIX: Flush then close

larsoner:
Previously, on `__exit__` we were effectively doing `fid.close()` then nothing (though we wanted to do `fid.flush()`), this corrects the order to be `fid.flush()` followed by `fid.close()`.

@drammock can you double-check the logic?

No test to update because this doesn't have an obvious deterministic effect AFAIK. But I'm hoping it might solve #327, where some lines seem to be missing in the `.tab` file. But either way this can be merged as a bugfix.


larsoner:
You're right, order should be fixed now.

drammock:
thanks @larsoner

PR #336, opened 2017-09-19
MRG: Add bias

larsoner:
Adds bias measure to d' calculations.

Ready for review/merge from my end.

### DIFF
--- a/expyfun/analyze/_analyze.py
+++ b/expyfun/analyze/_analyze.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 """Analysis functions (mostly for psychophysics data).
 """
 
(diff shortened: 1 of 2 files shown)


rkmaddox:
LGTM

rkmaddox:
Worth noting (here, but not in the code since you specify the exact calculation) that "bias" has a few different meanings. Sometimes it's as we say here (my preference), sometimes it's a ratio (can't remember exactly of what to what) where 1 means unbiased, and sometimes it's the log of that ratio. It can also be reported in the stimulus units, whereas here it is normalized. So care should be taken when writing about it, and reading about it.

larsoner:
@rkmaddox I can add a note along the lines that there are multiple defs, that's useful for people to think about

@drammock do you have time to look in the next week or so?

PR #339, opened 2017-11-29
FIX: Fix recon

larsoner:
@rkmaddox @maddycapp27 because trackers are identified by their memory location, it is actually possible for tracker names to repeat in the unlucky case where the same memory address is used by multiple trackers (e.g., one is instantiated after another is garbage collected). I hit such a use case recently. This fixes the reconstruction.

Should we also fix the naming, e.g. by also tagging with the time of instantiation (to the millisecond), too?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,5 @@ env/
 # Gedit autosave files
 *~
 
+# pytest
+.cache
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # caution: testing won't work on windows, see README
 
 PYTHON ?= python
-NOSETESTS ?= nosetests
+PYTEST ?= pytest
 CTAGS ?= ctags
 
 all: clean inplace test test-doc
@@ -42,12 +42,12 @@ inplace:
 
 nosetests:
 	rm -f .coverage
-	$(NOSETESTS) expyfun
+	$(PYTEST) expyfun
 
 test: clean nosetests flake
 
 test-doc:
-	$(NOSETESTS) --with-doctest --doctest-tests --doctest-extension=rst doc/
+	$(PYTEST) --doctest-modules --doctest-ignore-import-errors --doctest-glob='*.rst' ./doc/
 
 version:
 	@expr substr `git rev-parse HEAD` 1 7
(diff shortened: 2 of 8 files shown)


larsoner:
@maddycapp27 any thoughts on this one?

maddycapp27:
@larsoner Sorry for the delay! I don't see any harm in tagging with more info but you would be a better judge. Could the time tag replace the ID number or would it be another data line?

rkmaddox:
If time is only recorded to the millisecond, then we still ned the other id. It's entirely possible that two trackers could be created during the same millisecond.

larsoner:
I'm thinking both time and id

larsoner:
Turns out `time.time()` is good to the microsecond, but I still made it use `id-time` now instead of just `id`.

Okay with you @rkmaddox @maddycapp27 ?

maddycapp27:
Sounds good to me

larsoner:
@drammock do you have time to look?

drammock:
I can review later this morning.

drammock:
tests pass in both py2 and py3 on my Linux, and both `tracker_dealer.py` and `tracker_dealer_doublesided.py` examples work.  Only thing that sticks out at me is that the example scripts don't end with a `plt.show()`.  The tracker IDs in the verbose output are annoyingly long to identify by eye, but at least they won't collide now.

larsoner:
Thanks for the quick look @drammock

rkmaddox:
Not to beat a dead horse, but if we wanted to ensure uniqueness, but also make them more readable, we could hash the id-time tag.

larsoner:
+1 by me once someone is motivated to do it (i.e., not a high priority)

rkmaddox:
For whomever does it in the future:

```python
import hashlib

hashlib.md5('foo').hexdigest()
# outputs 'acbd18db4cc2f85cedef654fccc4a4d8'
hashlib.md5('fool').hexdigest()
# outputs 'dd7b696b96434d2bf07b34f9c125d51d'
```

drammock:
I thought of that too, but I didn't think that md5 hashes were much of an
improvement in readability... probably somewhat quicker to make the "are
these two different" judgement at a glance, but still hard to make the "is
this the same one I was looking at several seconds ago?" judgement.

rkmaddox:
Agree, though it would at least allow eyeballing that wouldn't depend on the last few characters.

If we are OK with one in 4.3 billion chance of collision: https://github.com/zacharyvoase/humanhash.

PR #341, opened 2017-12-07
FIX: Reenable parallel port triggering

larsoner:
@mfreilly can you try this branch on the experimental computer that will run the auditory paradigm? Once it appears to work, we should probably verify that the timing accuracy (jitter and delay) with an oscilloscope.

### DIFF
--- a/doc/python_reference.rst
+++ b/doc/python_reference.rst
@@ -30,6 +30,7 @@ Classes:
 
    ExperimentController
    EyelinkController
+   ParallelTrigger
    TDTController
 
 Functions:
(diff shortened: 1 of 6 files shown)


larsoner:
@rkmaddox you might also be interested in testing this to move toward TDT independence

rkmaddox:
Noted. Though our SPDIF solution is already in the works.

larsoner:
Yes this would be more for if you wanted triggering during the sound stimulus (or without sound playing), e.g. to give times of visual stim flips or response windows, etc.

rkmaddox:
Ahhh gotcha. So I could potentially use just channel 0 for sound stimulus, and then do the others with parallel port. Not a bad idea.

larsoner:
If you want to spend $20 on a parallel port card and do some testing, feel free :)

PR #346, opened 2018-04-12
MRG: Minor fixes

larsoner:
Some minor updates.

### DIFF
--- a/expyfun/io/_parse.py
+++ b/expyfun/io/_parse.py
@@ -107,7 +107,7 @@ def read_tab(fname, group_start='trial_id', group_end='trial_ok',
         header.append(header.pop(header.index(group_end)))
         b2s = np.where([line[1] == group_end for line in lines])[0]
     if len(b1s) != len(b2s) or not np.all(b1s < b2s):
-        raise RuntimeError('bad bounds in %s:\n{0}\n{1}'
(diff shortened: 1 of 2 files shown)


expyfun/io/_parse.py
```diff
@@ -107,7 +107,8 @@ def read_tab(fname, group_start='trial_id', group_end='trial_ok',
         header.append(header.pop(header.index(group_end)))
         b2s = np.where([line[1] == group_end for line in lines])[0]
     if len(b1s) != len(b2s) or not np.all(b1s < b2s):
-        raise RuntimeError('bad bounds:\n{0}\n{1}'.format(b1s, b2s))
+        raise RuntimeError('bad bounds in %s:\n{0}\n{1}'
+                           .format(fname, b1s, b2s))
```
drammock: this won't work. `print('%s: {0} {1}'.format('foo', 'bar', 'baz'))` yields `%s: foo bar` so you're variable `b2s` is never getting printed.

drammock: should be: `raise RuntimeError('bad bounds in {0}:\n{1}\n{2}'.format(fname, b1s, b2s))`

larsoner:
Thanks @drammock, all better now

PR #353, opened 2018-06-18
MRG: Text input

maddycapp27:
This is a function to let participants type responses (and view those responses as they type).

To do:
handling of special characters (like 'space', punctuation)
blinking cursor?
tests

To debate:
How to exit input? ``stop_key`` parameter v. a submit button to click (former is easier for participant, but more prone to prematurely exiting)
Recording response -- is logging by ``ec.screen_text()`` sufficient? annoying? Should the function return the final string?

### DIFF
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -1149,7 +1149,7 @@ class ExperimentController(object):
                 elif letter == 'space':
                     text = text.ljust(len(text) + 1)
                 elif letter in letters:
-                        text += letter
+                    text += letter
                 self.screen_text(text + '|', pos=pos, color=color,
                                  font_name=font_name, font_size=font_size,
                                  wrap=wrap, units=units, log_data=False)
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -15,6 +15,7 @@ import warnings
 from os import path as op
 from functools import partial
 import traceback as tb
+import string
 
 import numpy as np
 
@@ -423,7 +424,8 @@ class ExperimentController(object):
 
 # ############################### SCREEN METHODS ##############################
     def screen_text(self, text, pos=[0, 0], color='white', font_name='Arial',
-                    font_size=24, wrap=True, units='norm', attr=True):
+                    font_size=24, wrap=True, units='norm', attr=True,
+                    log_data=True):
         """Show some text on the screen.
 
         Parameters
@@ -450,6 +452,8 @@ class ExperimentController(object):
             Should the text be interpreted with pyglet's ``decode_attributed``
             method? This allows inline formatting for text color, e.g.,
             ``'This is {color (255, 0, 0, 255)}red text'``.
+        log_data : bool
+            Whether or not to write a line in the log file.
 
         Returns
         -------
@@ -463,8 +467,9 @@ class ExperimentController(object):
         scr_txt = Text(self, text, pos, color, font_name, font_size,
                        wrap=wrap, units=units, attr=attr)
         scr_txt.draw()
-        self.call_on_next_flip(partial(self.write_data_line, 'screen_text',
-                                       text))
+        if log_data:
+            self.call_on_next_flip(partial(self.write_data_line, 'screen_text',
+                                           text))
         return scr_txt
 
     def screen_prompt(self, text, max_wait=np.inf, min_wait=0, live_keys=None,
@@ -1102,13 +1107,14 @@ class ExperimentController(object):
         self._response_handler.check_force_quit()
 
     def text_input(self, stop_key='return', pos=[0, 0], color='white',
-                   font_name='Arial', font_size=24, wrap=False, units='norm',
+                   font_name='Arial', font_size=24, wrap=True, units='norm',
                    all_caps=True):
         """Allows participant to input text and view on the screen.
         
         Parameters
         ----------
         """
+        letters = string.ascii_letters
         stop = False
         text = str('')
         while not stop:
@@ -1124,10 +1130,11 @@ class ExperimentController(object):
                 else:
                     if all_caps:
                         letter = str.upper(letter)
-                    text += letter
+                    if letter in letters:
+                        text += letter
                 self.screen_text(text + '|', pos=pos, color=color,
                                  font_name=font_name, font_size=font_size,
-                                 wrap=wrap, units=units)
+                                 wrap=wrap, units=units, log_data=False)
             self.flip()
         
 
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -1134,8 +1134,8 @@ class ExperimentController(object):
         all_caps : bool
             Whether the text should be displayed in all caps.
         """
-        letters = string.ascii_letters
-        text = str('')
+        letters = string.ascii_letters + ' '
+        text = str()
         while True:
             letter = self.wait_one_press(timestamp=False)
             if letter == stop_key:
@@ -1146,7 +1146,7 @@ class ExperimentController(object):
             else:
                 letter = ' ' if letter == 'space' else letter
                 letter = letter.upper() if all_caps else letter
-                text += letter if letter in letters or letter == ' ' else ''
+                text += letter if letter in letters else ''
             self.screen_text(text + '|', pos=pos, color=color,
                              font_name=font_name, font_size=font_size,
                              wrap=wrap, units=units, log_data=False)
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -1148,9 +1148,8 @@ class ExperimentController(object):
                 elif letter == 'space':
                     text = text.ljust(len(text) + 1)
                 else:
-                    if all_caps:
+                    if all_caps and letter in letters:
                         letter = str.upper(letter)
-                    if letter in letters:
                         text += letter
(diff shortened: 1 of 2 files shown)


drammock:
possibly useful?  https://bitbucket.org/pyglet/pyglet/src/86e656f7154c32aaae42f594dcc73b86f6b3e6bd/examples/text_input.py#lines-30:32

rkmaddox:
I definitely vote for a key press to finish input. People are used to using Enter for that, and requiring a mouse would be pretty annoying.

Since people will be typing, we will need to worry about accidental quit key presses. What we might want for that is to allow a mouse button press for quitting. If not that, then a quit key combo.

Function should both return the final string and also write the final string to the tab file. `ec.screen_text()` is going to write its string every time a key is pressed and that string is updated, which will get pretty messy. If there is a way to only write when the submit button is pressed that would be optimal I think?

rkmaddox:
Or I guess one other way to do it would be to not listen for quit keys while a text box is open. That might be cleaner, but it requires pressing the submit button before one can quit.

rkmaddox:
Would also add an option `all_caps=True` since people won't be using the shift button.

maddycapp27:
> we will need to worry about accidental quit key presses

As far as I know, ``ec.screen_text()`` doesn't check for force quit. From outside the booth it would be easy enough to hit the ``stop_key`` to exit text input and then the force quit key to quit.

> going to write its string every time a key is pressed and that string is updated, which will get pretty messy

Agreed, potentially solved by adding an argument to ``ec.screen_text()`` of ``write_data=True`` and then set it false for ``ec.text_input``?

maddycapp27:
Any thoughts on the CircleCI build immediately failing?

expyfun/_experiment_controller.py
```diff
@@ -1101,6 +1106,58 @@ def check_force_quit(self):
...
+            else:
+                if letter == 'backspace':
+                    text = text[:-1]
+                elif letter == 'space':
+                    text = text.ljust(len(text) + 1)
+                else:
+                    if all_caps:
+                        letter = str.upper(letter)
+                    if letter in letters:
+                        text += letter
```
drammock: you could reduce one level of nested `ifs` like this:
```
while not stop:
    letter = self.wait_one_press(timestamp=False)
    if all_caps and letter in letters:
        letter = letter.upper()
    if letter == stop_key:
        stop = True
    else:
        if letter == 'backspace':
            text = text[:-1]
        elif letter = 'space':
            text = text + ' '
        elif letter in letters:
            text += letter
```

larsoner:
(note the status link https://circleci.com/gh/maddycapp27/expyfun/185 is nested under your user/fork rather than the `gh/LABSN/expyfun/*` path)

larsoner:
Now I have no idea why CircleCI isn't building your PR :(

expyfun/tests/test_experiment_controller.py
```diff
@@ -439,6 +439,15 @@ def test_button_presses_and_window_size():
...
         assert_equal(out[0], '1')
+        fake_button_press(ec, 'a', 0.3)
+        fake_button_press(ec, 'return', 0.5)
+        ec.text_input()
+        fake_button_press(ec, 'a', 0.3)
+        fake_button_press(ec, 'space', 0.35)
+        fake_button_press(ec, 'backspace', 0.4)
+        fake_button_press(ec, 'comma', 0.45)
+        fake_button_press(ec, 'return', 0.5)
+        ec.text_input()
```
larsoner: and here, `assert ec.text_input() == 'a,'`

maddycapp27:
I thought I just deleted it for me (I got rid of the webhook in my fork settings)?

Clearly I know just enough to be dangerous :(

expyfun/_experiment_controller.py
```diff
@@ -1101,6 +1106,57 @@ def check_force_quit(self):
...
+                elif letter == 'space':
+                    text = text.ljust(len(text) + 1)
+                else:
+                    if all_caps and letter in letters:
+                        letter = str.upper(letter)
+                        text += letter
+                self.screen_text(text + '|', pos=pos, color=color,
+                                 font_name=font_name, font_size=font_size,
+                                 wrap=wrap, units=units, log_data=False)
+            self.flip()
```
larsoner: might as well have it return `text`

expyfun/tests/test_experiment_controller.py
```diff
@@ -439,6 +439,15 @@ def test_button_presses_and_window_size():
         fake_button_press(ec, '1', 0.3)
         out = ec.wait_for_presses(1.5, live_keys=['1'], timestamp=False)
         assert_equal(out[0], '1')
+        fake_button_press(ec, 'a', 0.3)
+        fake_button_press(ec, 'return', 0.5)
+        ec.text_input()
```
larsoner: then here you can write `assert ec.text_input() == 'a'`

expyfun/_experiment_controller.py
maddycapp27: I like that

expyfun/_experiment_controller.py
```diff
@@ -1101,6 +1106,57 @@ def check_force_quit(self):
...
+            else:
+                if letter == 'backspace':
+                    text = text[:-1]
+                elif letter == 'space':
+                    text = text.ljust(len(text) + 1)
+                elif letter in letters:
+                    text += letter
+                self.screen_text(text + '|', pos=pos, color=color,
+                                 font_name=font_name, font_size=font_size,
+                                 wrap=wrap, units=units, log_data=False)
```
larsoner: Can this be simplified to:
```
while True:
    letter = self.wait_one_press(timestamp=False)
    if letter == stop_key:
        break
    if letter == 'backspace':
        text = text[:-1]
    else:
        letter = ' ' if letter == 'space' else letter.upper()
        letter = letter.upper() if all_caps else letter
        text += letter
    self.screen_text(...)
    self.flip()
```
?

larsoner: (One change that might matter is that here I have removed the check for `letter in letters` because e.g. `'-'.upper() == '-'`)

drammock: ooh, I like that better.

maddycapp27: the space is an issue because it just reads as an empty string

drammock: wait.  the `break` means that `self.flip()` won't get called when `stop_key` is pressed, and the text will just stay there on screen.

maddycapp27: Nevermind, apparently that's not an issue anymore?

drammock: easy enough to solve though:
```
...
if letter == stop_key:
    self.flip()
    break
...
```

larsoner: sure, add it there, add an extra branch level, or add it to the end after the `while ` loop is over

larsoner: Ahh so Pyglet doesn't give characters, it gives human-readable names. Bummer

maddycapp27:
Any other changes? @larsoner @drammock @rkmaddox

expyfun/tests/test_experiment_controller.py
```diff
@@ -439,6 +439,15 @@ def test_button_presses_and_window_size():
         fake_button_press(ec, '1', 0.3)
         out = ec.wait_for_presses(1.5, live_keys=['1'], timestamp=False)
         assert_equal(out[0], '1')
+        fake_button_press(ec, 'a', 0.3)
+        fake_button_press(ec, 'return', 0.5)
+        assert_equal(ec.text_input(), 'A')
```
larsoner: FYI we no longer need to do `assert_equal`, with `pytest` you can do `assert a == b` directly and it does good things automagically

expyfun/_experiment_controller.py
```diff
@@ -1101,6 +1106,54 @@ def check_force_quit(self):
...
+            letter = self.wait_one_press(timestamp=False)
+            if letter == stop_key:
+                self.flip()
+                break
+            if letter == 'backspace':
+                text = text[:-1]
+            else:
+                letter = ' ' if letter == 'space' else letter
+                letter = letter.upper() if all_caps else letter
+                text += letter if letter in letters or letter == ' ' else ''
```
drammock: optional: this might be a little clearer if, up above:
`letters = string.ascii_letters + ' '`
and then this becomes
`text += letter if letter in letters else ''`

expyfun/_experiment_controller.py
```diff
@@ -1101,6 +1106,54 @@ def check_force_quit(self):
...
+            for multi-line text. Inappropriate for text requiring
+            precise positioning or centering.
+        units : str
+            Units for `pos`. See `check_units` for options. Applies to
+            `pos` but not `font_size`.
+        all_caps : bool
+            Whether the text should be displayed in all caps.
+        """
+        letters = string.ascii_letters
+        text = str('')
```
drammock: this is a tad redundant; I'd do `text = ''` or `text = str()`

expyfun/_experiment_controller.py
maddycapp27: makes sense

larsoner:
Thanks @maddycapp27

larsoner:
FYI it renders fine

http://labsn.github.io/expyfun/generated/expyfun.ExperimentController.html#expyfun.ExperimentController.text_input

drammock:
oops,  never added Returns to the docstring after making it return `text`

PR #354, opened 2018-06-21
FIX: Logging for Text Input

maddycapp27:
I forgot to make sure the logging actually happened! Also fixed the docs.

### DIFF
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -1133,6 +1133,11 @@ class ExperimentController(object):
             `pos` but not `font_size`.
         all_caps : bool
             Whether the text should be displayed in all caps.
+
+        Returns
+        -------
+        text : str
+            The final input string.
         """
         letters = string.ascii_letters + ' '
         text = str()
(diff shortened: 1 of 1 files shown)


expyfun/_experiment_controller.py
```diff
@@ -1151,6 +1156,8 @@ def text_input(self, stop_key='return', pos=[0, 0], color='white',
                              font_name=font_name, font_size=font_size,
                              wrap=wrap, units=units, log_data=False)
             self.flip()
+        self.call_on_next_flip(partial(self.write_data_line, 'text_input',
+                                       text))
```
larsoner: Why wait until `call_on_next_flip` rather than immediately do it (i.e., call `self.write_data_line(...)` directly)?

maddycapp27: I copied from ``ec.screen_text()`` assuming there was a reason I didn't know about. Doing it directly makes more sense to me.

larsoner: In the case of `ec.screen_text` the text does not actually get displayed until the next call to `.flip()`. Here the text has already been obtained/completed

larsoner:
Thanks @maddycapp27

PR #359, opened 2018-07-10
WIP: Calibration Interrupt

maddycapp27:
I'd attempted this before and it was surprisingly messy so guidance/debate is welcome. @larsoner @drammock @rkmaddox 

The goal is to be able to re-calibrate during an experiment when a given key is pressed.

Things to consider:
1. ``ec`` has the button press info but ``el`` needs to be called -- so which should the function live in? I'd previously gotten it to "work" with the function living in ``ec`` and importing ``el`` to ``ec`` after both were initialized but it was pretty gross.
2. Which functions (if any) should automatically check for re-calibration? We'd want to be able to re-calibrate before a trial if the subject isn't able to get fixation (and then probably abort the rest of the trial after calibration) and also between trials (so we could re-calibrate after a break).
3. Any key press functions in input controllers like ``_retrieve_keyboard_events`` or ``_correct_presses`` will need to recognize the calibration key.

### DIFF
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -1105,6 +1105,9 @@ class ExperimentController(object):
     def check_force_quit(self):
         """Check to see if any force quit keys were pressed."""
         self._response_handler.check_force_quit()
+    
+    def check_calibration(self):
+        """Check if calibration interrupt keys were pressed."""
 
     def text_input(self, stop_key='return', pos=[0, 0], color='white',
                    font_name='Arial', font_size=24, wrap=True, units='norm',


maddycapp27:
@larsoner It looks like circle build is working

larsoner:
> Which functions (if any) should automatically check for re-calibration? We'd want to be able to re-calibrate before a trial if the subject isn't able to get fixation (and then probably abort the rest of the trial after calibration) and also between trials (so we could re-calibrate after a break).

I don't think any should, I think you should make a function to do this.

It should probably be a method of `EyelinkController`, because it already contains a reference to its parent `ec`, so it can call the proper `ec` polling function.

> Any key press functions in input controllers like _retrieve_keyboard_events or _correct_presses will need to recognize the calibration key.

This sounds too messy, as it requires burdening generic press-event code with paradigm-specific checks. A cleaner design is probably to know that there is some interval during which a recalibration can be requested, and press the button then.

maddycapp27:
Have not tested this with an actual eyelink connected yet and have not added tests.

I've started with what I think the minimum scope of ``ec.on_every_wait()`` should be -- it's only included in ``wait_secs`` in _utils. To my knowledge this gets called for ``ec.wait_secs()`` and at the beginning of functions that have a ``min_wait`` parameter. The next thing is whether we want wait functions to be called by ``ec.screen_prompt()``, ``ec.wait_one_press()``, etc. The only other place I can think to include it would be during ``el.wait_for_fix()`` but doing so would call for aborting a trial and probably should be done on an experiment by experiment basis.

rkmaddox:
My feeling is that putting the functions into the base waiting function (which I think is `wait_secs` ?) is the way to do it. That way every function that waits at all, including the ones you listed, calls the `on_every_wait` functions at some interval.

rkmaddox:
Ahh, I see what you're saying. Functions like `wait_one_press` don't use `wait_secs`, so they'll also require modification. That's a tough one. I think those should also do the `on_every_wait` functions? Or maybe have a flag that defaults to `True`? Is there ever a reason we _wouldn't_ want them to run?

maddycapp27:
Yes, ``wait_secs`` is called during those functions in ``_init_wait_click`` or ``_init_wait_press`` for the duration of ``min_wait`` but then the remainder of the waiting is controlled by a while loop that depends on whether there has been a click/a live key pressed and the ``max_wait``.

Upshot, it should check at the beginning of those functions (but actually doesn't in my test script).

maddycapp27:
Another question is whether the recalibration should show the calibration prompt or not. Since the default behavior of ``el.calibrate()`` is to show the prompt, that is what I have for the moment.

@larsoner @drammock Any concerns so far?

maddycapp27:
Tested with real eyelink and seems to behave

expyfun/_experiment_controller.py
```diff
@@ -680,6 +681,31 @@ def call_on_every_flip(self, function):
...
+            The function to call. If ``None``, all the "on every flip"
+            functions will be cleared.
+
+        See Also
+        --------
+        ExperimentController.call_on_next_flip
+
+        Notes
+        -----
+        See `flip_and_play` for order of operations. Can be called multiple
```
larsoner: This does not look correct / relevant

expyfun/_eyelink_controller.py
```diff
@@ -137,14 +137,17 @@ class EyelinkController(object):
...
 
     Notes
     -----
     The data will be saved to the ExperimentController ``output_dir``.
     If this was `None`, data will be saved to the current working dir.
     """
     @verbose_dec
-    def __init__(self, ec, link='default', fs=1000, verbose=None):
+    def __init__(self, ec, link='default', fs=1000, verbose=None,
+                 calibration_key=['c']):
```
larsoner: Do not use mutable default arguments. Tuple is immutable so `('c',)` is better here

expyfun/tests/test_eyelink_controller.py
```diff
@@ -46,6 +47,9 @@ def test_eyelink_methods():
         # run much of the calibration code, but don't *actually* do it
         el._fake_calibration = True
         el.calibrate(beep=False, prompt=False)
+        fake_button_press(ec, 'c')
+        el.check_recalibrate(prompt=False)
```
larsoner: Shouldn't you assert that this is True ?

maddycapp27: The test gets hung up on the prompt screen. Could put in a ``fake_button_press`` if you want to test the prompt?

larsoner: I don't understand. It currently gets stuck, or will get stuck with my suggested addition of an `assert ` in this line preceding the `el...`?

maddycapp27: Read too quickly, you are correct

expyfun/_experiment_controller.py
```diff
@@ -680,6 +681,20 @@ def call_on_every_flip(self, function):
         else:
             self._on_every_flip = []
 
+    def call_on_every_wait(self, function):
+        """Add a function to be executed on every wait.
+
+        Parameters
+        ----------
+        function : function | None
+            The function to call. If ``None``, all the "on every flip"
```
drammock: shouldn't this say `all the "on every wait"`?

expyfun/_eyelink_controller.py
```diff
@@ -137,14 +137,17 @@ class EyelinkController(object):
         Sample rate to use. Must be one of [250, 500, 1000, 2000].
     verbose : bool, str, int, or None
         If not None, override default verbose level (see expyfun.verbose).
+    calbration_keys : list
+        Keys that will trigger recalibration when check_recalibration.
```
drammock: can you expand this docstring line a bit?  The reference to `check_calibration` needs context (i.e., it's not a parameter of this constructor, so explain what it is / where it lives)...  is it referring to the new `check_recalibrate` method?  (if so, the method name is wrong here)

drammock: also, you say it should be `list` but you initiate it with a tuple.  Will it work with any iterable?  if so, say `iterable` instead of `list`

expyfun/_eyelink_controller.py
```diff
@@ -381,6 +385,36 @@ def calibrate(self, beep=False, prompt=True):
         self._start_recording()
         return fname
 
+    def check_recalibrate(self, keys=None, prompt=True):
+        """Compare key buffer to recalibration keys and calibrate if matched.
+
+        This function always uses the keyboard, so is part of abstraction.
```
drammock: Not sure I understand ths note.

expyfun/_eyelink_controller.py
```diff
@@ -381,6 +385,36 @@ def calibrate(self, beep=False, prompt=True):
...
+
+        Parameters
+        ----------
+        keys : list or string
+            keys to check if prompt recalibration
```
drammock: keys to check against the (set of) key(s) that trigger calibration.

expyfun/_eyelink_controller.py
```diff
@@ -381,6 +385,36 @@ def calibrate(self, beep=False, prompt=True):
...
+        prompt : bool
+            Whether to show the calibration prompt or not
```
drammock: Whether to show the calibration prompt screen before starting the calibration procedure

PR #360, opened 2018-07-10
MRG: Text input more user friendly

maddycapp27:
This alters ``ec.text_input()`` such that brief instructions can be given and so that the cursor appears before the participant begins typing.

### DIFF
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -1106,7 +1106,7 @@ class ExperimentController(object):
         """Check to see if any force quit keys were pressed."""
         self._response_handler.check_force_quit()
 
-    def text_input(self, stop_key='return', instruction_string='Type '
+    def text_input(self, stop_key='return', instruction_string='Type'
                    ' response below', pos=[0, 0], color='white',
                    font_name='Arial', font_size=24,  wrap=True, units='norm',
                    all_caps=True):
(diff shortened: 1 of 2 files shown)


maddycapp27:
@larsoner @drammock ready for review. Thanks much!

Travis fail looks related but it's only 1 of 4 and I can't seem to recreate the error on my machine.

expyfun/_experiment_controller.py
```diff
@@ -1106,15 +1106,18 @@ def check_force_quit(self):
         """Check to see if any force quit keys were pressed."""
         self._response_handler.check_force_quit()
 
-    def text_input(self, stop_key='return', pos=[0, 0], color='white',
-                   font_name='Arial', font_size=24, wrap=True, units='norm',
+    def text_input(self, stop_key='return', instruction_string='Type '
+                   ' response below', pos=[0, 0], color='white',
+                   font_name='Arial', font_size=24,  wrap=True, units='norm',
```
drammock: FYI, we normally add new keyword arguments to the *end* of the function signature (for backwards compatibility with user scripts that were passing args positionally).  In this case it's probably fine since `text_input` is so new, but now you know.

drammock: Also, `instruction_string` appears to have two spaces between `Type` and `response`.

maddycapp27: Ooh good catch.

maddycapp27:
I can definitely work on an example but don't want to hold up our undergrads (both of whom are working on/piloting experiments using this function)

maddycapp27:
Tests passing now -- @rkmaddox and I came to a consensus on what we wanted the new functionality to be before I opened the PR so I'm assuming he's good to merge.

rkmaddox:
+1 for merge

larsoner:
Thanks @maddycapp27

Issue #358, opened 2018-06-28
Many problems with video presentation


rkmaddox:
So my labbies and I are having a heck of a time getting video to work. Here is a summary of the issues as I understand them, though they will fill in the rest.

0. AVBin will not play quicktime movies (we can deal with that, will just switch to x264)
1. AVBin freezes near (but not quite at) then end of our video (x264 encoded)
2. AVBin has strange image quality (Sara will post screen cap). To me it looks like a missed keyframe, but it never fixes itself. When the video is played in a media player it looks fine.

Those are the problems with trying to play a movie in pyglet. Strangely, the issues don't appear in the `simple_video` example.

Another, bigger, issue, which @drammock alluded to in #273 is that the sync may be off when playing videos. What we really need is access to the video frames and the ability to get the textures and flip as needed. This would also solve the problem of getting it in sync with the audio and triggers. Basically, we would handle video as we did in 1v2a, but instead of saucers it would be video frames. We have tried very hard to pull textures from videos in pyglet, but it is not working for us yet.

Labbies are currently signing up for github and will post more info here.

rkmaddox:
This will also be essential to figure out in order to present video on the monitor with the corresponding audio from the TDT.

drammock:
For the record, I never got it satisfactorily working in my initial attempts, so I'm not surprised you're struggling.  I bailed when I realized I didn't actually need AV synchrony, and used a hacky solution of hiding the expyfun window and popping up a borderless full-screen VLC window to play the video.  That doesn't mean we can't get this to work, but it will probably require some additions to / overhaul of the video code.

rkmaddox:
For the time sync, would [providing our own clock](http://pyglet.readthedocs.io/en/pyglet-1.3-maintenance/programming_guide/time.html#user-defined-clocks) with t=0 as the start time help?

rkmaddox:
@drammock Yeah, it is pretty much essential for us now. We have many upcoming experiments where playing natural videos is the only way to go, and they have to be time synced.

rkmaddox:
The sync itself should _not_ be difficult. It was essentially solved with 1v2a. What we need to be able to do is just use a pyglet video source as a container for a sequence of frames, pull the frames out one by one, and flip them at the time we deem correct.

sfiscell:
This is an image of the video quality: 
[...]

PR #361, opened 2018-07-13
WIP: Faster image updates

larsoner:
Needs sanity checks


larsoner:
Closing since I think we should avoid getting more dependent on Pyglet. I think this was meant to address #358 (maybe). @rkmaddox did you need this? Did you find a workaround for video yet?

rkmaddox:
We found a workaround, though I don't remember what it was. I agree that we should not become more dependent on Pyglet, especially since it won't be able to handle the multi-channel audio.

PR #362, opened 2018-09-06
ENH: Refocus on Windows

larsoner:
The procedure to test in case someone is curious, only on Windows (for now):

1. On `master`, run `sync_test.py` with `response_device='keyboard'`:
    1. Press any key (like space) and notice the flashing speeds up (because it does not wait the 0.5 sec.) 
    2. Press alt-tab to switch to any other process
    3. Press any key (like space) and notice the flashing does not change, and pressing "ctrl" does not close the window.
    4. Click in the expyfun window
    5. Press "ctrl" to close the window (it closes).
2. On this branch, repeat the above, but steps (iv) is not necessary to get step (v) behavior -- it happens already on step (iii) (i.e., focus is regained automatically).


larsoner:
This has been used successfully in th wild, so I'll merge.

PR #364, opened 2019-01-30
ENH: Move to pytest syntax

larsoner:
Long-overdue migration to `pytest` syntax.

Also bumps Travis to 3.7.


larsoner:
Coverage only hurts because I removed some code lines (I think), so I'll merge once CIs are happy

PR #367, opened 2019-04-08
WIP: fix pyglet screen num (wants screen instance)

drammock:
pyglet 1.3.2 is complaining when I pass an int `screen_num` to ExperimentController (or when I set `SCREEN_NUM` in the machine config).  This fixes that.  Not well tested yet, but have a look.
@larsoner

### DIFF
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -249,9 +249,9 @@ class ExperimentController(object):
                 screen_num = int(get_config('SCREEN_NUM', '0'))
             if monitor is None:
                 import pyglet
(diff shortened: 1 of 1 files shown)


larsoner:
Looks reasonable but I haven't checked how this interacts with the monitor stuff above these changes

larsoner:
Looks like `monitor` isn't even in the described arguments / parameters list. I wonder if we should just remove it from the `__init__`. It's code from 5 years ago (!) and I doubt we need it / use it. But maybe for now I'll try documenting it properly.

In the meantime this fixes a bug. It would be good to have a test but I'll add one in a follow-up PR

PR #369, opened 2019-04-09
fix: class 'Raw' doesn't work with current pyeparse

drammock:
At some point pyeparse __init__ used to include a line `Raw = RawEDF` (introduced in [this commit](https://github.com/pyeparse/pyeparse/commit/c7f61e84dbad49163fbd3fade103ebf0db700330#diff-51ad22576f4f2ea6d270fed13a3badd3R7) but not sure when it was removed). Anyway, since it was removed the creation of a graph at the end of the pupillometry code block no longer works. This one-liner fixes it.


larsoner:
Thanks @drammock

PR #381, opened 2019-06-05
FIX: fix typo

maddycapp27:
fix a typo in rtmixer docs

### DIFF
--- a/doc/rtmixer_installation.rst
+++ b/doc/rtmixer_installation.rst
@@ -20,7 +20,7 @@ Here are some stub installation instructions for each OS:
 
        $ sudo apt install libportaudio2  # only if on Linux!
        $ brew install portaudio  # only if on OSX!
-       $ pip install python-sounddevice
+       $ pip install sounddevice
        $ python -m sounddevice  # just to see if it worked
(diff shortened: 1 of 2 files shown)


drammock:
There's another typo on line 593 of experiment controller, backgound -> background, which is causing Travis to fail. Can you fix that too? (I'm writing from phone so I can't do it myself at the moment)

drammock:
@larsoner looks like we're still running [a travis test in a py2.7 environment](https://travis-ci.org/LABSN/expyfun/jobs/541865658), and anything in `expyfun` that depends on `mne` is now failing.  Do we need to continue supporting 2.7 for some reason?  I thought it was the EyeLink `pylink` module, but they have py3k versions of that now...

larsoner:
Probably no need to support 2.7 anymore, especially given our versioning system. Okay with you @rkmaddox @maddycapp27 to go 3.5+ only?

maddycapp27:
I'm fine with it --  my computer and the research computers are either transitioned or imminently will transition. @rkmaddox Here's some more motivation to update your computer :)

larsoner:
Thanks @maddycapp27

PR #386, opened 2019-07-18
MRG: Add multi-channel support, fixed delay

### LINKED ISSUE #383: n-channel audio where n > 2
The very good news is that the `rtmixer` audio device is talking to our RME sound card with these settings:

```
audio_controller=dict(TYPE='sound_card', 
                      SOUND_CARD_API='ASIO',
                      SOUND_CARD_BACKEND='rtmixer',
                      SOUND_CARD_NAME='ASIO HDSPe FX',
                      SOUND_CARD_FS=48000)
```

So that is excellent. The trouble I am having now is that there are a lot of checks in there that try to force the sound to be two channels. I have removed enough of them that I don't get an error when I give `load_buffer` six channels of audio, but it still only plays from the first two channels.

There are still a lot of situations where all those two-channel checks (and doing things like transposing a stim that is, say, 1000 x 2). Namely: whenever headphones are in use. Should we keep all those checks and manipulations but condition them on `audio_device` itself having two channels? Should we have a `stereo` flag? Or should we just no longer assume two channels and remove the checks? I think one of the first two options makes the most sense.

@larsoner do you have any >2 channel audio interfaces you can use to test? Possible your MOBO soundcard has multiple outs. Or I could send an old M-Audio USB sound card I have that probably has 4 outputs. I can check when I get home tonight.

larsoner:
Haven't tested the multichannel stuff or fixed delay on this branch, but in theory it should work since fixed delay seemed to work with what is now `SOUND_CARD_FIXED_DELAY` added to the time. For my built in sound card on Linux, 10ms (`0.01`) is enough.

@rkmaddox if you are bored feel free to try it, but it might need a bit of work still since I haven't re-tested with a scope or tried my multichannel card, but at least with my `pulse` ALSA "card" (with 32 outputs for some reason) this works in `sync_test.py`:
```
with ExperimentController('SyncTest', full_screen=True, noise_db=-np.inf,
                          participant='s', session='0', output_dir=None,
                          suppress_resamp=True, check_rms=None,
                          version='dev', n_channels=8) as ec:
    ec.load_buffer(np.tile(np.r_[0.1, np.zeros(99)][np.newaxis], (8, 1)))  # RMS == 0.01
```
and the messages are reasonable:
```
2019-07-17 22:53:54,900 - INFO    - Expyfun: Using version 2.0.0.dev0 (requested dev)
2019-07-17 22:53:55,498 - INFO    - Expyfun: Setting up sound card audio using rtmixer backend and 8 channels
2019-07-17 22:53:55,548 - INFO    - Expyfun: using sound card 'pulse' (devices[19]) via 'ALSA', 8 channels @ 44100 Hz, 8.7 ms latency
2019-07-17 22:53:55,548 - INFO    - Expyfun: Variable audio delay
```

Closes #383.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -235,7 +235,7 @@ trim_doctests_flags = True
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
-    'numpy': ('https://www.numpy.org/devdocs', None),
+    'numpy': ('https://numpy.org/devdocs', None),
     'scipy': ('https://scipy.github.io/devdocs', None),
     'matplotlib': ('https://matplotlib.org', None),
     'sklearn': ('https://scikit-learn.org/stable', None),
--- a/doc/getting_started.rst
+++ b/doc/getting_started.rst
@@ -26,9 +26,11 @@ expyfun requires several libraries for full functionality:
   - ``matplotlib``
   - ``pyglet``: 1.2.0 or later is recommended
 
-- Optional Python libraries:
+- Optional libraries:
 
   - ``rtmixer``: for high precision audio playback, see :ref:`rtmixer_installation`.
+  - ``pyparallel`` or ``inpout32.dll``: for parallel port triggering, see
+    :ref:`parallel_installation`.
   - ``TDTpy``: if using TDT on Windows
   - ``mne``: for filtering/resampling -- with CUDA if mne dependencies installed
   - ``pandas``: Some plotting functions
--- a/doc/parallel_installation.rst
+++ b/doc/parallel_installation.rst
@@ -32,10 +32,16 @@ Instructions differ between Linux and Windows:
     If you are on a modern Windows system (i.e., 64-bit), you'll need to:
 
     - Download the latest "binaries" archive from the `InpOut32 site`_
-    - Extract the **64-bit** file ``inpoutx64.dll`` from the ``.zip`` file
-    - Rename this file ``inpout32.dll``
-    - Place it in ``C:\Windows\System32\``
-    - Use the Device Manager to get the parallel port address, e.g.
-      ``'0x378'``, and set this as ``TRIGGER_ADDRESS`` in the config.
+    - Extract the files
+    - Run the ``Win32\InstallDriver.exe`` file (yes, even though it's in the
+      Win32 directory)
+    - Rename the **64-bit** file ``inpoutx64.dll`` to ``inpout32.dll``
+    - Place this file in ``C:\Windows\System32\``
+    - Use the Device Manager (or some other method) to get the parallel port
+      address (from Ports➡Properties➡Resources➡I/O Range), e.g. ``0x378``
+      or ``0xCFF4``, and set this as ``TRIGGER_ADDRESS`` in the config.
+    - If you have trouble, you can interactively test your parallel port using
+      the `parallel port tester application`_.
 
 .. _`InpOut32 site`: http://www.highrez.co.uk/downloads/inpout32/
+.. _`parallel port tester application`: https://www.downtowndougbrown.com/2013/06/parallel-port-tester/
--- a/doc/parallel_installation.rst
+++ b/doc/parallel_installation.rst
@@ -1,0 +1,41 @@
+:orphan:
+
+.. _parallel_installation:
+
+Parallel port triggering
+========================
+
+.. highlight:: console
+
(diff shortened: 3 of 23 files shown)


larsoner:
@rkmaddox @maddycapp27 on Linux with this config:
```
{
"SOUND_CARD_API": "ALSA",
"SOUND_CARD_BACKEND": "rtmixer",
"SOUND_CARD_FS": 44100,
"SOUND_CARD_NAME": "pulse",
"SOUND_CARD_FIXED_DELAY": 0.01,
"SOUND_CARD_NAME": "surround51",
"TRIGGER_ADDRESS": "/dev/parport0",
"TRIGGER_CONTROLLER": "parallel"
}
```
and setting
```
n_channels = 6  # my "surround51" device has 6 channels
```
I get < 1 ms jitter (probably < 100 us) and fixed ~11 ms delay. No idea why it's not exactly 10 ms but I'm not going to complain given its lack of jitter.

Hopefully the tiny extra parametrization of `sync_test.py` makes it easier to test on your rig.

rkmaddox:
We did a `sync_test`. We got about 2 ms jitter from click to monitor refresh. This is not ideal, but is also probably good enough. We got the same jitter with the Rift, and we were expecting worse, since there is some back and forth between the python script and Unity and we though that would add jitter but it doesn't seem to.

The 2 ms jitter seems to be because some variability in the monitor refresh time. And we are also getting some weird stuff with our monitor where at ostensibly 23 fps it is running at 23.8, at 59 it is running at 59.8, and at 60 I think it was like 60.3? No idea why that is happening, but it suggests the issue is not with the audio.

larsoner:
> We got about 2 ms jitter from click to monitor refresh.

Given your buffer length, this is what I would expect if the `start` was *not* working. Did you set `SOUND_CARD_FIXED_DELAY`? To see if it's having any effect, you should at least see the delay go up if you increase the value. In theory the jitter should go away, too (it does on Linux ALSA at least).

> The 2 ms jitter seems to be because some variability in the monitor refresh time.

I guess it could be due to this. It's probably worth getting that $10 parallel port card installed to have a more definitive "time I told it to play" trigger.

rkmaddox:
We had the `SOUND_CARD_FIXED_DELAY` to 25 ms, and we are using a 64-sample buffer. OK I'll buy the parallel port card.

larsoner:
Updated instructions once you get your card:

https://551-11614571-gh.circle-artifacts.com/0/html/parallel_installation.html#parallel-installation

larsoner:
FYI for my NVIDIA GPU on a Windows machine w/built in sound I had some settings in the "NVIDIA Control Panel" I could configure (which I did just for the `python.exe` rather than globally), such as:

1. disabling triple buffering
2. enabling vsync (just to be safe, Pyglet should do this during config anyway)
3. setting the maximum number of pre-rendered frames to 1

This seemed to get me to very low jitter levels (< 2 ms somewhere), but some of the frames just completely "missed" (the flashes did not happen) so there is something wacky going on. I was even just using WASAPI and the built-in laptop sound card -- I could get the sound and video to overlap nicely with a `'SOUND_CARD_FIXED_DELAY': 0.02` value, at least when the flashes actually showed up:

image

This is the sort of thing that would maybe be fixed by moving to GLFW or some other library for window management and rendering. But in the meantime it's worth looking into your settings @rkmaddox to see if tweaking some of these helps, and what I've found suggests that things should hopefully work!

drammock:
I also encountered the "some flashes just don't happen" problem when testing at UMN in April. Hopefully I'll get out there again sometime soon so we can get another datapoint / so I can pitch in on the troubleshooting. Hard to do much Windows testing from my linux laptop :/

larsoner:
@rkmaddox @maddycapp27 I've done a bit of testing on a native Windows machine with a parallel port, and updated the instructions for parallel port usage accordingly.

It looks like using `asio4all` and a built-in sound card I can get a +/-2ms jitter relative to the video (and parallel port trigger), which IIRC is what you saw on your systems. The jitter between the video and parallel port was reliable and < 1 ms. Given the time it takes to actually draw a frame, and the fact that we will add support for augmented-sound-channel-based triggering (giving us precise timing the sound started) this seems like a workable solution. WDYT?

FYI @drammock on my Windows machine with a proper NVIDIA card and the built-in Windows drivers, I didn't get any missed / non-white frames. Going to try installing the latest NVIDIA driver now, though, so I can eventually test the driver settings.

larsoner:
Okay, in it goes. Let's iterate further in subsequent PRs if necessary

PR #394, opened 2019-09-04
MRG, ENH: Use sound card as triggering device

larsoner:
By setting:
```
"SOUND_CARD_TRIGGER_CHANNELS": 2,
```
This makes it so that any `load_buffer(sound_stimulus)` call where `sound_stimulus` has shape `(n_ch, n_samples)` will prepend two channels with impulses at the first sample (a one followed by zeros) and prepare it to be played (sent to the sound card as shape `(n_ch + n_stim_channels, n_samples)`.

I've only tested with 2 (or 0) stim channels and 2 playback channels and it seems to do what I'd expect here. @rkmaddox @maddycapp27 can you try it on your system and see if 1) it actually produces a trigger and 2) the jitter is zero?

Not 100% sure you want a one followed by zeros, if you need some other triggering scheme / value then we'll have to figure out how to get you the value you need API-wise and implementation-wise (conversion betweer whatever `int` value you need coming out of the digital channels and the `float32` dtype that we use natively).


larsoner:
@drammock feel free to review/merge when you get a chance

larsoner:
Reverted the "addition trick" with justification in the code.

I'll go ahead and merge this once the CIs come back happy.

PR #395, opened 2019-09-10
WIP: Add PTB backend

larsoner:


larsoner:
Closing since I think we won't end up going this route

PR #396, opened 2019-09-11
ENH: Log stamping

larsoner:
Adds to the `.log` file that we attempted to stamp some triggers.


larsoner:
@drammock can you review and merge if you're happy? I think this is a pretty straightforward enhancement

expyfun/_experiment_controller.py
```diff
@@ -1931,11 +1933,10 @@ def _stamp_binary_id(self, id_, delay=0.03, wait_for_last=True):
         if not isinstance(id_, (list, tuple, np.ndarray)):
             raise TypeError('id must be array-like')
         id_ = np.array(id_)
-        if not np.all(np.logical_or(id_ == 1, id_ == 0)):
+        if not np.all((id_ == 1) | (id_ == 0)):
```
drammock:
```suggestion
        if not np.all(np.in1d(id_, [0, 1])):
```

drammock: I think this is equivalent and maybe slightly clearer.  The current version might be carelessly misread as checking that `id_` is either all 1s or all 0s.

larsoner:
Yes to both of your points. I figure if people want to control the trigger duration and separation separately they can, but assume YAGNI until we hit a use case.

PR #398, opened 2019-09-14
FIX: Don't stop stimuli too early or (usually) wait for the stop

larsoner:


larsoner:
This is good incremental progress so I'll go ahead and merge

PR #399, opened 2019-09-21
MRG, ENH: Joystick control and Pyglet 1.4

### LINKED ISSUE #385: MAINT: Support pyglet 1.4
There are some nasty backward incompatible bits, and the docs have not been updated to reflect changes in API :(

larsoner:
This should work in principle but Pyglet appears to have a [nasty limitation](https://pyglet.readthedocs.io/en/stable/programming_guide/input.html) (see also [PsychoPy's note](https://www.psychopy.org/api/hardware/joystick.html#joystick-pyglet-and-pygame)):

> Please note that you need a running application event loop for the joystick button an axis values to be properly updated

I plan to look into why this is the case. It might require some updates to Pyglet.

The alternative would be to swap out our backend entirely to GLFW, which would probably / maybe be a lot more work.

In the meantime I hope to have made things Pyglet 1.4+ compatible, too. At least the experiment runs locally (except for the joystick actually working) and the tests pass!

- [x] Migrate video code
- [x] Migrate avbin->FFmpeg in Circle
- [x] FFmpeg Travis
- [x] FFmpeg Azure
- [x] FFmpeg AppVeyor
- [x] Remove `expyfun._utils.wait_secs` or make private; update `ExperimentController.wait_secs` to use `self._dispatch_events`
- [x] Get Joysticks working by mimicking the Pyglet event loop behavior
- [x] Test on Windows and Linux
- [x] Add tests using `monkeypatch`

Closes #385.


larsoner:
Seems to be working:

joyExp-2019-09-22-14-33-15

larsoner:
And here is the example that gives rise to that image (joystick + user interaction necessary):

https://639-11614571-gh.circle-artifacts.com/0/html/auto_examples/experiments/joystick_experiment.html#sphx-glr-auto-examples-experiments-joystick-experiment-py

@drammock do you have time to look and merge? If not we can probably rely on the tests here.

Unfortunately there is no real way I can think of to test `Joystick` support at the moment without a physical one. Maybe we can find some way to hack it at some point like we do for `ParallelTrigger` where we actually get some coverage by allowing a "null" controller, but I'm not sure the tests would give us too much information in that case, since most of the code is Pyglet-joystick-specific :(

larsoner:
- Verified it works on Linux (with no changes!)
- Tweaked a variable name
- Added tests (hooray `monkeypatch` from `pytest` making life easy!)

larsoner:
I've rebased now that #400 is in

drammock:
I had a skim through the changes but don't really have time for a proper review ATM.  If you're comfortable with the tests you've written, then +1 for merge.

larsoner:
I think it should be good, we can iterate with @rkmaddox and @maddycapp27 if it does not work properly

PR #400, opened 2019-09-26
FIX: Missing triggers

larsoner:
I don't love this fix/workaround but if there is something fishy with the stream time provided by the RME ASIO drivers, this should make sure we don't inadvertently cut off triggers.

Also removes a `mixer.start_time = mixer.time` assignment that is just leftover cruft.


larsoner:
@maddycapp27 reports that this does fix the problem, so I'll go ahead and merge this

PR #402, opened 2019-09-27
FIX: Bump up tolerance for drift

larsoner:
Even when using the same clock 10 µs can be tough with Python overhead. So let's bump our tol up to 50.

cc @maddycapp27 let me know if this works for you. If it does I'll leave open for a bit in case there are other joystick-related fixes we need.


larsoner:
Reported as working so I'll merge

PR #406, opened 2020-01-28
ENH: Add SOUND_CARD_TRIGGER_INSERTION

larsoner:
Allows `"SOUND_CARD_TRIGGER_INSERTION": "append"` instead of the default `"prepend"`.


larsoner:
Reported by end-users as working. @drammock did you want to take a quick look?

drammock:
looks reasonable to me. Checking my understanding of the code: this allows the extra sound card channel that's being used for triggers to come after the audio channels instead of before, right?

larsoner:
Yes indeed

drammock:
tx @larsoner

rkmaddox:
Thanks!

PR #408, opened 2020-02-11
MRG, ENH: Speed up buffering

### LINKED ISSUE #407: ec.load_buffer() is slow
Loading a fairly large stimulus to the buffer (53 chans, 16 sec at 48kHz) is taking longer than expected: 6.93 sec, 7.04 sec, and 7.00 sec on three test runs.

Ross mentioned that this may be a known bug related to unnecessary memory copying.

larsoner:
@JustinTFleming can you try adding this to your script after EC initialization:
```
ec.set_stim_db(60.)  # assuming you have stim_rms=0.01, the default value
assert ec._stim_scaler == 1.
```
and then see if 1) stimuli load faster than 7 sec, and 2) that they still sound correct? The assertion is necessary to make sure we avoid a scaling later.

This PR should eliminate all copies except the last one (concatenating the sound and the trigger values), there triggering a copy which would be difficult to avoid (we would have to tell `rtmixer` to play two sounds at the same time and hope it did so correctly, which is dicey).

If this ends up being the bottleneck, we could make a new triggering mode where you have to supply the sound stimulus *with the trigger channels zeroed out* (e.g., for 6 channel + 1 trigger channel, you would already supply an array of shape `(7, n_samples`). But let's see how fast we are already with these changes.

Closes #407

### DIFF
--- a/expyfun/tests/test_experiment_controller.py
+++ b/expyfun/tests/test_experiment_controller.py
@@ -290,6 +290,9 @@ def test_ec(ac, hide_window):
             ec.load_buffer(click)
         with pytest.warns(UserWarning, match='exceeds stated'):
             ec.load_buffer(noise)
+        if ac != 'tdt':  # too many samples there
+            with pytest.warns(UserWarning, match='slow'):
+                ec.load_buffer(np.zeros(10000001, dtype=np.float32))
 
         ec.stop()
         ec.set_visible()
--- a/expyfun/tests/test_experiment_controller.py
+++ b/expyfun/tests/test_experiment_controller.py
@@ -10,6 +10,7 @@ import pytest
 from numpy.testing import assert_allclose
 
 from expyfun import ExperimentController, visual
+from expyfun._experiment_controller import _get_dev_db
 from expyfun._utils import (_TempDir, fake_button_press, _check_skip_backend,
                             fake_mouse_click, requires_opengl21,
                             _wait_secs as wait_secs, known_config_types)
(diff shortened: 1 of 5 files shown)


JustinTFleming:
I'm getting an `AssertionError` at that new line. I didn't have stim_rms set at all in the `ExperimentController` call, but the error is the same with any combination of `stim_db` or `stim_rms` set (or omitted) in the ec call.

Also of note: Working from one of Maddy's scripts, I also added a missing `check_rms='wholefile'` argument, which cut buffer time by an order of magnitude; it's now 700-800 ms, in line with performance in her code with stimuli of the same duration.

larsoner:
> I'm getting an AssertionError at that new line. I didn't have stim_rms set at all in the ExperimentController call, but the error is the same with any combination of stim_db or stim_rms set (or omitted) in the ec call.

What is the value of `ec._stim_scaler`? If you change the line to this it should spit it out on failure:
```
assert ec._stim_scaler == 1., ec._stim_scaler
```

> Working from one of Maddy's scripts, I also added a missing check_rms='wholefile' argument, which cut buffer time by an order of magnitude; it's now 700-800 ms, in line with performance in her code with stimuli of the same duration.

Makes sense it would be the RMS checking. Good to know. I hadn't profiled to see if it would be the memcopies but probably should have:
```
In [1]: import numpy as np                                                                                                                                                          
In [2]: n_ch, dur, fs = 63, 16, 48000                                                                                                                                               
In [3]: n_samp = dur * fs
In [4]: data = np.random.RandomState(0).randn(n_samp, n_ch)                                                                                                                         
In [5]: %timeit data.copy()                                                                                                                                                         
55.8 ms ± 739 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
That's pretty quick but nonetheless it would still be nice to include these performance enhancements if possible, though. This scales linearly with time, so for example a 16 minute stimulus would take 3 sec per copy. (It also takes many GB of memory so not too feasible, but still.)

larsoner:
Also pushed a commit to:

1. Warn if you are using `windowed` mode and a lot of samples
2. Speed up the `windowed` mode by an order of magnitude

larsoner:
Okay we might as well get this in, @drammock can you look and merge if you're happy?

expyfun/tests/test_experiment_controller.py
```diff
@@ -52,6 +53,30 @@ def test_unit_conversions(hide_window, ws):
...
+            samples_out = ec._validate_audio(samples_in)
+            assert samples_out.shape == (1000, 2)
+            assert samples_out.dtype == np.float32
+            assert samples_out is not samples_in
+        for order in 'CF':
+            samples_in = np.zeros((2, 1000), dtype=np.float32, order=order)
+            samples_out = ec._validate_audio(samples_in)
+            assert samples_out.shape == samples_in.shape[::-1]
+            assert samples_out.dtype == np.float32
+            # transpose that we do will not copy just make a view
```
drammock: this comment is hard to parse, can you re-word it or add punctuation to make it easier to understand?

expyfun/tests/test_experiment_controller.py
```diff
@@ -52,6 +53,30 @@ def test_unit_conversions(hide_window, ws):
...
+            assert samples_out.base is samples_in
+            samples_out.flags['WRITEABLE'] = True  # undo our read-only safety
+            samples_out[:] = 1
+            assert_allclose(samples_in.T, samples_out)
```
drammock: I don't get why this passes; `samples_in` is all zeros, and you just changed `samples_out` to all ones.  Are you just proving a point that `samples_out` is a view?  Why do we need to test that?

expyfun/tests/test_experiment_controller.py
```diff
@@ -265,6 +290,9 @@ def test_ec(ac, hide_window):
             ec.load_buffer(click)
         with pytest.warns(UserWarning, match='exceeds stated'):
             ec.load_buffer(noise)
+        if ac != 'tdt':  # too many samples there
+            with pytest.warns(UserWarning, match='slow'):
```
drammock: match string seems too generic.  How about `many samples is slow`?

expyfun/tests/test_experiment_controller.py
larsoner: > Are you just proving a point that samples_out is a view? Why do we need to test that?

Yes exactly. The idea is that, if copies are slow, we should keep the number of copies we make to a minimum. But I think the `samples_out.base is samples_in` should already ensure this, so I'll remove these lines.

rkmaddox:
Thank you!

PR #409, opened 2020-03-02
MAINT: Remove workaround

larsoner:
This workaround is 5 years old, hopefully we can get rid of it now.

Also adds a bit of maybe useful debugging information.

### DIFF
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -914,7 +914,7 @@ class ExperimentController(object):
         platform_event_loop.stop()
 
 # ############################### OPENGL METHODS ##############################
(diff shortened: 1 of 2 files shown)


larsoner:
@drammock do you remember what made it so that we needed this workaround?

drammock:
I'm not positive, but it might have been an issue with left-aligning a multiline message?

-------- Original Message --------

larsoner:
Okay, let's try removing it and see if we end up with problems. With the current implementation and modern Pyglet it actually causes some problems (left aligning with the extra spaces does not work properly). And the workaround is 5 years old, so hopefully we can do without it nowadays...

PR #412, opened 2020-05-12
MRG, ENH: Add SNR and spectral estimation example

larsoner:
@drammock can you make sure I didn't do anything silly here? Just want to give people a primer on stimulus power, etc.


larsoner:
Rendered example:

https://729-11614571-gh.circle-artifacts.com/0/html/auto_examples/stimuli/stimulus_power.html#sphx-glr-auto-examples-stimuli-stimulus-power-py

Going to push commits to try to make CIs happy, as seems necessary every few months with shifting deps/pyglet stuff...

PR #413, opened 2020-06-05
BUG: Fix video playback with other visuals

### LINKED ISSUE #358: Many problems with video presentation

larsoner:
Closes #358

### DIFF
--- a/examples/stimuli/advanced_video.py
+++ b/examples/stimuli/advanced_video.py
@@ -28,6 +28,7 @@ with ExperimentController(**ec_args) as ec:
     all_presses = list()
     fix = visual.FixationDot(ec)
     text = None  # don't draw at first
+    screenshot = None  # don't have one yet
     ec.load_video(movie_path)
     ec.video.set_scale('fill')
     ec.screen_prompt('press 1 during video to toggle pause.', max_wait=1.)
@@ -38,6 +39,10 @@ with ExperimentController(**ec_args) as ec:
         ec.video.draw()
         if text is not None:
             text.draw()
+            if screenshot is None:
+                screenshot = ec.screenshot()
+                if building_doc:
+                    break
         fix.draw()
         fliptime = ec.flip()
         presses = ec.get_presses(live_keys=[1], relative_to=t_zero)
@@ -67,9 +72,6 @@ with ExperimentController(**ec_args) as ec:
                 else:
                     ec.video.set_visible(True)
                     ec.video.play()
-        screenshot = ec.screenshot()
-        if building_doc:
-            break
     ec.delete_video()
     preamble = 'press times:' if len(all_presses) else 'no presses'
     msg = ', '.join(['{0:.3f}'.format(x[1]) for x in all_presses])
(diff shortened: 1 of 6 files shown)


larsoner:
@drammock this fixes a bug with video playback when other visuals are used. Can you give it a look?

larsoner:
@LiesbethGijbels can you see if the `advanced_video.py` example works for you? In theory it should allow you to use `expyfun` video functions to show your video rather than `cv2`.

larsoner:
@drammock can you look and try this out?

drammock:
Not easily.  see spatialaudio/python-rtmixer#22

drammock:
running `advanced_video.py` example on Linux looks good after installing `pillow`(which is an unstated dependency, maybe only for screenshots?).  I see the video shrinking and then moving rightward, with the background color changing behind it.  Keyboard button presses do not work to pause the video however.

drammock:
After setting config value `RESPONSE_DEVICE` to `keyboard`, I still can't pause the video.

larsoner:
I broke that bit of the example, it should be fixed now (pausing using 1. pauses things on my machine). If you get no keyboard response at all, perhaps try `simple_experiment.py` to see if other experiments work or not. It will tell us if the problem is specific to this example, or to `expyfun`/pyglet overall

drammock:
hmm, button presses not working in simple_experiment either. I've probably forgotten some dumb aspect of the setup; it's been so long since I set up a new system like this.

larsoner:
Make sure the window that pops up actually has focus, you might need to click on it to get it to really have focus

drammock:
> you might need to click on it to get it to really have focus

yeah, I've been doing that; still no dice.

larsoner:
Maybe try changing Pyglet versions? `pip install pyglet==1.3.2` or `pyglet<1.5` to get some 1.4 variant...?

If that doesn't work, which window manager are you running? I can try to replicate. Running just from a terminal with `python -i whatever.py` I assume?

drammock:
downgrading to pyglet 1.4 worked!

PR #416, opened 2020-06-23
MRG, ENH: Add move_mouse_to

larsoner:
Tested on Linux and macOS, Windows in progress. @drammock feel free to review and if you're happy I'll merge once it works on Windows.

cc @LiesbethGijbels this will allow you just to use `ec.move_mouse_to((0, 0))` instead of using `pyautogui`.

### DIFF
--- a/expyfun/_input_controllers.py
+++ b/expyfun/_input_controllers.py
@@ -554,7 +554,7 @@ class Mouse(object):
             point = POINT()
             point.x = x
             point.y = y
-            _user32.ClientToScreen(self.ec._hwnd, byref(point))
+            _user32.ClientToScreen(self.ec.window._hwnd, byref(point))
             # Move the mouse
             _user32.SetCursorPos(point.x, point.y)
         else:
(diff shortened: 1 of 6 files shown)


larsoner:
Scratch that, seems to work fine on Windows already. No mouse moving works on Windows 7 for me (not even pyautogui) but this code works fine on my Windows 10 install.

expyfun/_input_controllers.py
```diff
@@ -513,6 +515,57 @@ def _point_in_tri(self, pos, tri):
...
+            # Convert from window to global
+            from pyglet.window.win32 import POINT, _user32, byref
+            point = POINT()
+            point.x = x
+            point.y = y
+            _user32.ClientToScreen(self.ec.window._hwnd, byref(point))
+            # Move the mouse
+            _user32.SetCursorPos(point.x, point.y)
+        else:
+            # https://stackoverflow.com/questions/2433447/how-to-set-mouse-cursor-position-in-c-on-linux/2433488  # noqa
```
drammock, line 561: FYI if you use the "share" link on stackoverflow questions or answers, you get stable short URLs that don't contain the whole question title. Or you can just omit everything after the question number (the final digits are logged-in user IDs, so they can track your link-sharing behavior) for an even shorter link:

```suggestion
            # https://stackoverflow.com/questions/2433447
```

PR #422, opened 2020-12-18
ENH: Allow parsing old files

larsoner:
Needed for some of @ktavabi's files


larsoner:
This was tested on 75 files and works pretty well and CIs are happy so I'll go ahead and merge, but others feel free to look

Issue #421, opened 2020-12-11
ENH: Using 2-triggers to account for EEG-sound card clock drift


rkmaddox:
We have recently discovered that the RME clock and the EEG clock run at ever so slightly different rates, slipping between 0.5 and 2 ms on a minute-long trial. There is a fairly simple fix, which is to use triggers to mark the beginning of the trial and another point or point(s) later in the trial.

This seems like a good use of the 2 trigger to us. A proposed solution would be that 2 triggers be sent at some configurable interval so their actual time compared to their nominal time tells us what the clock error is. A good default might be a 2 every 10 s. Trials less that 10 seconds are only looking at .2 ms accumulated error, so not so bad. If you were doing really long trials (say 60 s) and you wanted only one 2, you could set the interval to 50 s or something. There seems to be very little cost to anyone who doesn't have this problem of having an extra set of "timing triggers" in their recordings.

Thoughts?

larsoner:
I don't see a problem with adding an option to the sound card triggering scheme to insert 2 triggers. I would say that the default should be just to add a single 2 to the end of the stimulus, no matter what its duration of the stimulus actually is. I have a feeling that this drift is going to be fixed across systems, so this can probably be estimated once during a given recording (likely EEG-sound-card pairing, but per recording to be safe) and having a 2 at the end of the stimulus will be enough to do this with accuracy down to the microsecond probably.

For related experiments and code, see https://github.com/mne-tools/mne-python/pull/8540. The TL;DR is that Neuromag MEG and EGI EEG systems recording simultaneously have this problem, and computing a linear drift term fixed over the recording seems to fix things so far.

For experiments that want to be extra careful (or for early testing), we can have a `ec.set_sound_card_drift_interval(10.)` to set it to something other than "start and end". Sound good?

rkmaddox:
We could to that. It makes things a bit more complicated when there are variable-length stimuli, but perhaps we could jsut have one trigger time that can take either a number for seconds or a text input `'end'`? It also would need to be a short period of time before the end of the stim, because the trigger has some duration.

Strangely, it is _not_ fixed across systems, but it is stable over a recording. Our same pieces of equipment have shown different drifts over time, with a sudden change on one day that persisted. I think this is because we removed a piece of equipment that the RME had been sitting on, which very slightly reduced its temperature, affecting the resonance of the quartz crystal.

rkmaddox:
[...]

PR #424, opened 2021-04-28
Use 2-triggers to account for EEG-sound card clock drift.

tomstoll:
I made an attempt at inserting 2-triggers so we can use them to estimate/correct clock drift. Currently, the default behavior is `'end'` which inserts a single 2-trigger at the very end of the trial. It can also accept a single time (in seconds) or a list of times when the 2-triggers should be stamped. Positive values will be the time from start of trial and negative values will be time from end of trial. I think an empty list would result in no 2-triggers being stamped. It still needs error handling to make sure triggers don't overlap or occur outside the trial time.


expyfun/_sound_controllers/_sound_controller.py
```diff
@@ -103,6 +109,10 @@ def __init__(self, params, stim_fs, n_channels=2, trigger_duration=0.01,
         self._id_after_onset = (
             str(params['SOUND_CARD_TRIGGER_ID_AFTER_ONSET']).lower() == 'true')
         self._extra_onset_triggers = list()
+        if np.isscalar(params['SOUND_CARD_DRIFT_TRIGGER']):
+            self._drift_trigger_time = [params['SOUND_CARD_DRIFT_TRIGGER']]
+        else:
+            self._drift_trigger_time = list(params['SOUND_CARD_DRIFT_TRIGGER'])
```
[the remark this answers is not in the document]
tomstoll: I like this, I'll try to put it in. I just need to add `drift_trigger = params['SOUND_CARD_DRIFT_TRIGGER']` before the first line here and modify the code further down slightly, right? In the assertion, should we also allow numpy arrays?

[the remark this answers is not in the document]
larsoner: > I just need to add drift_trigger = params['SOUND_CARD_DRIFT_TRIGGER'] before the first line here and modify the code further down slightly, right?

Yeah I think so. I didn't test the above code so it might need some other adjustments, too

>  In the assertion, should we also allow numpy arrays?

No I wouldn't, it's easy enough to have people do `list(my_array)` and it simplifies code at our end

tomstoll:
Tried to implement the above suggestion and add some warnings in case triggers overlap or are outside the stimulus window.
We also thought it would be good to use the bitwise or operation instead of adding the triggers in case any overlap, but because the triggers are floats by this point I wasn't able to do that. Any thoughts on this @larsoner?

larsoner:
> We also thought it would be good to use the bitwise or operation instead of adding the triggers in case any overlap, but because the triggers are floats by this point I wasn't able to do that. Any thoughts on this @larsoner?

We could make `_make_digital_trigger` return `int32` data, and add a new `_scale_digital_trigger` that does the `((trigs << 8) * self._trig_scale).astype(np.float32)` step. Then you can add in bit-perfect-int-land and scale afterward.

rkmaddox:
That's exactly the solution we discussed as well.

tomstoll:
Gave it a shot. I removed the lines that bitshift and convert to floats from `_make_digital_trigger` and changed from inserting the 2-triggers in `samples` to inserting them in `stim`. Then I added the `_scale_digital_trigger` function to do what was removed from `_make_digital_trigger`  and used that on `stim` after using `np.bitwise_or` to add in the 2-triggers. I also scale the triggers in `stamp_triggers` immediately after the call to `_make_digital_trigger`.

I also updated the warnings so that it will skip the triggers that occur outside the stimulus window and prevent an error, and checking for the overlapping 2-triggers actually works (previously always gave the warning because I forgot the `<trig2_len` part in the if statement).

larsoner:
Failure looks like it might be real and related:
```
expyfun\tests\test_experiment_controller.py:616: in test_sound_card_triggering
    ec.identify_trial(ttl_id=[1, 0], ec_id='')
expyfun\_experiment_controller.py:2038: in identify_trial
    self._id_call_dict[key](id_)
expyfun\_experiment_controller.py:2080: in _stamp_binary_id
    self._stamp_ttl_triggers(id_, wait_for_last, True)
expyfun\_experiment_controller.py:2121: in _stamp_ttl_triggers
    ids, wait_for_last=wait_for_last, is_trial_id=is_trial_id)
expyfun\_sound_controllers\_sound_controller.py:349: in stamp_triggers
    stim, ((0, 0), (0, self._n_channels)[self._stim_sl]), 'constant')
<__array_function__ internals>:6: in pad
    ???
c:\python37-x64\lib\site-packages\numpy\lib\arraypad.py:743: in pad
    pad_width = _as_pairs(pad_width, array.ndim, as_index=True)
c:\python37-x64\lib\site-packages\numpy\lib\arraypad.py:518: in _as_pairs
    return np.broadcast_to(x, (ndim, 2)).tolist()
<__array_function__ internals>:6: in broadcast_to
    ???
c:\python37-x64\lib\site-packages\numpy\lib\stride_tricks.py:411: in broadcast_to
    return _broadcast_to(array, shape, subok=subok, readonly=True)
c:\python37-x64\lib\site-packages\numpy\lib\stride_tricks.py:350: in _broadcast_to
    op_flags=['readonly'], itershape=shape, order='C')
E   ValueError: operands could not be broadcast together with remapped shapes [original->remapped]: (2,2)  and requested shape 
```

tomstoll:
Yup, been looking for it for a while. Think I got it. I wrote `triggers` where I should have had `stim`.

larsoner:
Now that #425 is merged, I did a merge with `upstream/main` -- your code was not tested on Linux or macOS because we ran out of Travis credits :(

Will merge assuming CIs come back happy, thanks @tomstoll !

expyfun/_sound_controllers/_sound_controller.py
```diff
@@ -246,17 +306,20 @@ def _make_digital_trigger(self, trigs, delay=None):
...
-        stim = np.zeros((n_samples, self._n_channels_stim), np.float32)
+        stim = np.zeros((n_samples, self._n_channels_stim), np.int32)
         offset = 0
         for trig in trigs:
             stim[offset:offset + n_on] = trig
             offset += n_each
         return stim
 
+    def _scale_digital_trigger(self, triggers):
+        return ((np.array(triggers, int) << 8) *
```
larsoner, line 320:
```suggestion
        return ((triggers << 8) *
```

larsoner:
... and I also pushed a tiny commit.

Just noticed that it's still entitled WIP -- did you want to check to make sure everything works locally or anything else, or is this good to go from your end?

larsoner:
> 53.48% of diff hit (target 95.00%)

This is not great, however -- maybe worth adding some tiny test that actually uses this feature?

tomstoll:
Just a couple more things to do on our end. We think it would be nice to put the samples when the 2-triggers occur in the tab file so they're easier to use during analysis. Looking at codecov, some of the changes in coverage it reports don't seem to be from the changes here, which I'm a little confused by.

larsoner:
Codecov is a little bit flakey, I would just concentrate on the changes here. For logging indeed it seems reasonable to log the relative times that will be stamped for each stimulus, probably just the list of float of times actually used would be good (`[1., 2., 3., 3.245]` or whatever).

tomstoll:
The times that 2-triggers are stamped will now be written to the tab file. I also tried to add some tests, so we'll see how that goes.

expyfun/tests/test_experiment_controller.py
```diff
@@ -620,6 +620,64 @@ def test_sound_card_triggering(hide_window):
         ec.load_buffer([1e-2])
         ec.start_stimulus()
         ec.stop()
+    # Test the drift triggers
+    with pytest.warns(UserWarning, match='Drift triggers overlap with onset '
+                      'triggers'):
```
larsoner, line 625: These context managers should be as targeted as possible, so should cover just the `load_buffer` step I think

tomstoll: Makes sense, will update.

tomstoll: Also removed WIP from the title, since I think we're pretty much good to go on our end.

expyfun/tests/test_experiment_controller.py
```diff
@@ -620,6 +620,65 @@ def test_sound_card_triggering(hide_window):
...
+        ec.stop()
+    audio_controller.update(SOUND_CARD_DRIFT_TRIGGER=[0.5, 0.501])
+    with ExperimentController(*std_args,
+                              audio_controller=audio_controller,
+                              trigger_controller='sound_card',
+                              n_channels=1,
+                              **std_kwargs) as ec:
+        ec.identify_trial(ttl_id=[1, 0], ec_id='')
+        with pytest.warns(UserWarning, match='Some 2-triggers overlap, times '
+                          'should be atleast 0.02 seconds apart.'):
```
larsoner, line 658: Error: expyfun/tests/test_experiment_controller.py:658: atleast ==> at least

tomstoll: Just fixed it in the tests and the warning. Should I wait for the checks to finish to see if there's anything else, or push it now?

larsoner:
I'd just push

expyfun/tests/test_experiment_controller.py
```diff
@@ -620,6 +620,65 @@ def test_sound_card_triggering(hide_window):
...
+        ec.stop()
+    audio_controller.update(SOUND_CARD_DRIFT_TRIGGER=[0.5, 0.501])
+    with ExperimentController(*std_args,
+                              audio_controller=audio_controller,
+                              trigger_controller='sound_card',
+                              n_channels=1,
+                              **std_kwargs) as ec:
+        ec.identify_trial(ttl_id=[1, 0], ec_id='')
+        with pytest.warns(UserWarning, match='Some 2-triggers overlap, times '
+                          'should be at least 0.02 seconds apart.'):
```
rkmaddox, line 658: This really depends on what the trigger length is set to be. But maybe this is fins a general suggestion, as there is no real reason why you might want the 2's to be any closer than that.

tomstoll: Right, just realized in `_sound_controller` I have it use `trig2time` for the suggestion, but I copied the actually error output for the test. Is that ok, or should I figure to make that adapt to the trigger length as well?

larsoner: `match` is a regex and we just want to make sure it emits some warning so I would just do `match='Some 2-triggers overlap.*'`

tomstoll: Don't know how to reference a different part of the code in a reply, but in `_sound_controller.py` I have

larsoner:
Awesome, thanks @tomstoll !

tomstoll:
Great, thanks for your help! This should close issue #421.

PR #428, opened 2021-05-27
fix install command

### LINKED ISSUE #427: DOC: bad install instructions
the quick setup instructions say to use conda to install pyglet, but that only works from conda forge.  Should add a `-c conda-forge` to the command, or else move `pyglet` to the pip install line.

cc @ChristinaZhao

drammock:
closes #427


larsoner:
Thanks @drammock

PR #430, opened 2021-07-09
ENH: Allow close-reopen cycles to work (maybe!)

### LINKED ISSUE #429: soundcard access error when re-running a script
When stopping and re-running a script in Spyder, I get a "device unavailable" error. On the first run, there is no sound card access issue.

larsoner:
Closes #429

I haven't tested this code locally because I need to set up a Spyder env. @rkmaddox @bonniekwlau feel free to try it in the meantime to see if it fixes your problem.

### DIFF
--- a/examples/sync/sync_test.py
+++ b/examples/sync/sync_test.py
@@ -40,8 +40,6 @@ from expyfun import ExperimentController, building_doc
 from expyfun.visual import Circle, Rectangle
 import expyfun.analyze as ea
 
-print(__doc__)
-
 n_channels = 2
 click_idx = [0]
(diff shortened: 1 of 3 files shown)


larsoner:
(or @drammock if you want to try to replicate the problem on `main` and then test you can also give it a shot)

larsoner:
Testing locally this makes it so that if you re-run a script, the first re-run still won't work but the second one will. I have no idea why this is the case, but it's at least better than on `main` where it won't ever work until you kill the kernel entirely. I'll go ahead and merge once the CIs come back happy so at least things get a little better!

PR #432, opened 2021-11-12
If sound card API not found, list options

buranconsult:
It might be helpful to list valid options to guide new users in getting expyfun set up.


larsoner:
Thanks @buranconsult !

PR #439, opened 2022-01-07
BUG: Fix video on Windows for Pyglet > 1.3

larsoner:
We had two problems, I think:

1. We need to call `pyglet.clock.tick(poll=True)` to get the video to advance, and 
2. The alpha for the video is messed up on Windows

(2) has been opened upstream as a Pyglet bug (https://github.com/pyglet/pyglet/issues/531), but the workaround here should be fine.

@JustinTFleming can you test it with latest pyglet (1.5.21) and pyglet-ffmpeg?

FWIW @drammock for me when I play the video it plays way too fast on Linux -- it should take ~6 sec but it takes ~1. But this seems to be the case with newer Pyglet even before these PRs. Do you want to take a look? IIRC you worked on the timing stuff originally...?

### DIFF
--- a/examples/stimuli/simple_video.py
+++ b/examples/stimuli/simple_video.py
@@ -33,6 +33,7 @@ with ExperimentController(**ec_args) as ec:
             screenshot = ec.screenshot()
         if building_doc:
             break
+        ec.check_force_quit()
     ec.delete_video()
     ec.flip()
     ec.screen_prompt('video over', max_wait=1.)
--- a/expyfun/_experiment_controller.py
+++ b/expyfun/_experiment_controller.py
@@ -890,13 +890,13 @@ class ExperimentController(object):
         self.video = None
 
 # ############################### PYGLET EVENTS ###############################
+# https://pyglet.readthedocs.io/en/latest/programming_guide/eventloop.html#dispatching-events-manually  # noqa
 
     def _setup_event_loop(self):
         from pyglet.app import platform_event_loop, event_loop
         event_loop.has_exit = False
         platform_event_loop.start()
         event_loop.dispatch_event('on_enter')
-
         event_loop.is_running = True
         self._extra_cleanup_fun.append(self._end_event_loop)
         # This is when Pyglet calls:
@@ -904,11 +904,12 @@ class ExperimentController(object):
         # which is a while loop with the contents of our dispatch_events.
 
     def _dispatch_events(self):
-        from pyglet.app import platform_event_loop
+        import pyglet
+        pyglet.clock.tick()
         self._win.dispatch_events()
         # timeout = self._event_loop.idle()
         timeout = 0
-        platform_event_loop.step(timeout)
+        pyglet.app.platform_event_loop.step(timeout)
 
     def _end_event_loop(self):
         from pyglet.app import platform_event_loop, event_loop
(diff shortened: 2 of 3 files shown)


JustinTFleming:
Just got a chance to test. The video does play, but it's clearly sped up for me too on Windows (playback time around 1s). 

For kicks I also tried it on another mp4 file and the same thing happens, with a bunch of dropped frames. Audio cuts out at this shortened video end time as well.

larsoner:
Okay we should probably fix the timing before bothering to merge this... maybe you should stick with 1.3 + AVBin until this part is solved, since I think the timing might work there...?

larsoner:
Okay this plus https://github.com/pyglet/pyglet/pull/533 fixes playback speed for me. Feel free to try it @JustinTFleming , hopefully the pyglet devs agree with my fix (or point out a better one!).

larsoner:
FWIW the "alpha on Windows" bug is fixed by https://github.com/pyglet/pyglet/pull/534, but this PR shouldn't hurt anything in the meantime.

@JustinTFleming if you update to latest `pyglet-1.5-maintenance` and (now that I'm merging this) and latest `expyfun`, video should work again for you on windows. It should also be smoother on Linux and macOS  / ffmpeg backend due to  pyglet/pyglet#533 !

JustinTFleming:
Yes, videos are now the correct duration on Windows as well, and `example-video.mp4` playback is much smoother!

With that working, I tried changing the video file in `simple_video.py`. Unfortunately, the video remains choppy / stop-and-start on test videos with the following characteristics (both 1920x1080 resolution):
- Video 1: 6-second mp4, 30 Hz fps, has audio, 3.5MB. Audio is also very choppy.
- Video 2: 30-second mp4, 30fps, no audio, 17.0MB.
I tried compressing those videos to m4v's to reduce file sizes, which got them down to 372KB and 3.74MB, respectively, but it didn't change video quality much in either case.

JustinTFleming:
Bit of guesswork here, not being very familiar with pyglet, but in `pyglet\media\codecs\__init__.py`, there's the following code to prepare the codecs. There's a comment that they should be listed in order, but on my machine wmf will be added ahead of ffmpeg (see below) -- possibly meaning it will be preferred? Also, have_ffmpeg() returns True, so pyglet is able to at least find that library.
`
def add_default_media_codecs():
    # Add all bundled codecs. These should be listed in order of
    # preference.  This is called automatically by pyglet.media.

    try:
        from . import wave
        add_decoders(wave)
        add_encoders(wave)
    except ImportError:
        pass

    if pyglet.compat_platform.startswith('linux'):
        try:
            from . import gstreamer
            add_decoders(gstreamer)
        except ImportError:
            pass

    try:
        if pyglet.compat_platform in ('win32', 'cygwin'):
            from pyglet.libs.win32.constants import WINDOWS_VISTA_OR_GREATER
            if WINDOWS_VISTA_OR_GREATER:  # Supports Vista and above.
                from . import wmf
                add_decoders(wmf)
    except ImportError:
        pass

    try:
        if have_ffmpeg():
            from . import ffmpeg
            add_decoders(ffmpeg)
    except ImportError:
        pass

    try:
        from . import pyogg
        add_decoders(pyogg)
    except ImportError:
        pass`

I can try moving the ffmpeg block ahead of wmf and see if that changes anything (for better or worse).

JustinTFleming:
Ooh, moving up the ffmpeg block makes a huge difference in playback quality -- drastically better. I think you were right @larsoner that it was defaulting to WMF!

I currently have no audio, but I'm testing at home without my usual RME setup, so I'll try this in the lab before getting alarmed about that.

larsoner:
Okay I opened https://github.com/pyglet/pyglet/pull/535, let's see what they say

PR #440, opened 2022-01-12
ENH: Use modern OpenGL everywhere

larsoner:
This at least moves us toward compatibility with pyglet-master by using modern OpenGL to do our post-flip draw. It actually doesn't yet generate proper visuals on pyglet `master`, TBD why. Probably some OpenGL state thing...

To merge this I think we need someone to have a system with reliable scope timing on `main` who can then test that things are still good on this branch. I can set this up someday on my proper Windows boot, but for now I'll just open this so we can have something to work from eventually.


NeuroLaunch:
I can work with you on this in the near future using our lab's system.

larsoner:
@drammock if you come back to this... really I think we should try to avoid any manual OpenGL calls if at all possible. I'd actually start from seeing if we could eliminate all of that stuff and use Pyglet to choose a double-buffered display etc. and do all our draws. Then check that the timing is good on that branch...

Issue #431, opened 2021-08-12
FIX: Expyfun package requirements


NeuroLaunch:
When updating the UW lab's expyfun environment I ran into an issue where a pyglet method was not found.  Specifically, `.glMatrixMode` was not recognized during a call to `exp_controller_module`.   Apparently the issue arises from changes in pyglet's support of OpenGL, whereby glMatrixMode and other methods have been deprecated.

@drammock has suggested specifying pyglet < 2.0 in the package setup (I had success downgrading to pyglet 1.5.18) or updating expyfun to support the recent versions of pyglet and OpenGL.

larsoner:
The current stable release is 1.5.18, and indeed 2.0-flavor (which are currently in prerelease phase) are not supported. We'll have to refactor drawing once 2.0.0 is actually released.

NeuroLaunch:
Sorry, I may have jumped the gun then.  I had run a 'conda update conda',
which may have changed the pyglet version; alternatively, I see our stim
computer has a clone of the pyglet repo so the 2.0 version may have been
pulled in.  (I'm still getting used to the expyfun coding environment.)

larsoner:
If using the pyglet repo you want to be on the 1.5-maintenance branch, not main/master (which has 2.0-dev)

larsoner:
@NeuroLaunch can you make a PR to update https://github.com/LABSN/expyfun/blob/main/doc/getting_started.rst to tell people to use `pyglet-1.5-maintenance` rather than `master` if they're going that route? This also got @JustinTFleming recently

NeuroLaunch:
@larsoner Yes, I can do that.

PR #442, opened 2022-01-20
Doc updates addressing #431

NeuroLaunch:
Mainly this PR adds recommendations in _getting\_started.rst_ to install a version of pyglet that works with the current expyfun codebase. I was unable to create a browser-friendly version of the document (i.e. via the `make` command), so I'm uncertain if the formatting is correct.  Also, for the current version of the document, as rendered in my brower, I see text like ":func:\`expyfun.get_config_path\`", which I think are broken links.  I'd be happy to make fixes with some guidance.

I also made minor additions to the _requirements\_test.yml_ and _.gitignore_ files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 *.orig
 .vscode
 doc/generated
+.DS_Store
 
 # C extensions
 *.so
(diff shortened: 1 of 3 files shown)


larsoner:
Thanks @NeuroLaunch !

PR #445, opened 2022-03-03
FIX: Fix bug with video playback

### LINKED ISSUE #444: Unable to change window size for video playback
Using the FFMPEG integration from the latest expyfun release, I am able to play MP4 videos at the size of the example video in simple_video.py: window_size = (720, 480). Playing at a smaller screen size and changing the aspect ratio (e.g. with window_size = (688, 387), which is 16:9) are both fine, but making the video window much larger -- or attempting to play it full screen -- causes problems on the two Windows machines I've tested.

As the screen size is increased to an intermediate level, e.g. (1050, 700) px, playback starts getting choppy, with what looks like dropped frames at the beginning. I can play the included example-video.mp4 up to full screen, but with similar issues. Trying to play other MP4 files at or near full screen leads to crashing, usually on the second ec.flip() call, with the error traceback below.

Curious if anyone has tried increasing the video window size on their machines and encountered similar issues.

-- PC info --
OS: Windows 10
Python version: 3.7
expyfun release: 6c11b8a
Running through Spyder (iPython 7.29.0)

Error stack from attempting to play the video with ec argument full_screen = True:

`Traceback (most recent call last):

  File "L:\Projects\Recovery_AV\simple_video_test.py", line 90, in <module>
    fliptime = ec.flip()

  File "L:\Projects\Recovery_AV\expyfun\_experiment_controller.py", line 1035, in flip
    function()

  File "L:\Projects\Recovery_AV\expyfun\visual\_visual.py", line 1308, in draw
    self._player.update_texture()

  File "C:\Users\jtf\.conda\envs\videoEnv\lib\site-packages\pyglet\media\player.py", line 461, in update_texture
    frame_rate = source.video_format.frame_rate

AttributeError: 'NoneType' object has no attribute 'video_format'`

larsoner:
I used a [4k 30fps](https://www.videezy.com/water/4162-hawaii-ocean-4k) and [1080p 30 fps](https://www.videezy.com/random-objects/5680-hot-air-balloon-at-night-free-4k-stock-video) on Linux with:
```
import expyfun
fnames = ['hawaii_blue_water_4K.mp4', '342572022.mp4']
with expyfun.ExperimentController('test', full_screen=True, version='dev',
                                  session='a', participant='b',
                                  output_dir=None) as ec:
    ec.load_video(fnames[0])
    ec.video.set_scale('fit')
    ec.video.play()
    while not ec.video.finished:
        if ec.video.playing:
            fliptime = ec.flip()
        ec.check_force_quit()
    ec.delete_video()
```
On Linux, on `main`:

1. With `fnames[0]` (4k) things are a slideshow (1-2 FPS) and I get the error reported in #444
2. With `fnames[1]` (1080p) things are smooth (probably the native 30 FPS) and there is no error

On this PR, for `[0]` I get the same playback but no error, and for `[1]` I get the same (good/correct) behavior as on `main`. And even when I set `set_scale(0.1)`, playback is still terrible for the 4k video, suggesting that it's a video decoding/resolution limitation rather than anything having to do with the window/display size on screen (again, at least on my Linux machine!).

@JustinTFleming can you try this branch and see if you observe the same behavior on Windows? (BTW, with bug reports it's usually best to come up with a minimal reproducible example like the above if possible -- it saves devs time and makes sure that everyone is comparing apples to apples, to the extent possible!)

If you do get the same behavior, then I would look into maybe making your videos have a slower framerate or resolution.
Closes #444


larsoner:
Still seems to work okay a couple of years later and tests are green so in it goes!

PR #447, opened 2022-10-27
ec.refocus sets window topmost

### LINKED ISSUE #446: ec.refoucs() doesn't bring window to the front
Calling `ec.refoucs()` grabs keyboard focus, but the window remains behind the IDE. Adding `_user32.SetWindowPos(m_hWnd, -1, 0, 0, 0, 0, 0x0001 | 0x0002)` to the `refocus` function solves the issue (I put it at line 2180 of _experiment_controller.py). Equivalent code is [already in the function](https://github.com/LABSN/expyfun/blob/6c11b8ac5d0b3c3576c1a93ee49708e8e5c68079/expyfun/_experiment_controller.py#L2182-L2185), but commented out. Is there a reason it's not used?

tomstoll:
Added a line in `experiment_controller.refocus()` which should bring the experiment window to the front and fix #446. Also removed commented out equivalent code from the same function. I quickly tested this on two computers running Windows 10 and one running Windows 7 and didn't notice any issues.


larsoner:
Argh the CIs haven't run in a while. I think we'll have to ignore / look carefully at the errors to see if they make any sense

tomstoll:
Ok. I still don't really know what I'm doing on GitHub, but I made sure to sync my fork with the current main branch before making the change in this PR. I'll look at the failed tests.

larsoner:
> I'll look at the failed tests.

They are almost certainly not related to your changes. Over time dependencies can have deprecations and CIs can change things slightly in ways that can change things. We haven't had a PR to `expyfun` in months, so these are going to cause CIs to fail.

I'll try to fix these first early next week, then we can merge `main` into this branch and the CIs will be meaningful. Until then, they probably aren't really.

expyfun/_experiment_controller.py
```diff
@@ -2182,13 +2182,10 @@ def refocus(self):
             m_hWnd = self._win._hwnd
             hCurWnd = _user32.GetForegroundWindow()
             if hCurWnd != m_hWnd:
+                _user32.SetWindowPos(m_hWnd, -1, 0, 0, 0, 0, 0x0001 | 0x0002)
```
larsoner, line 2185: Let's make explicit that we're using the same thing that was commented out before
```suggestion
                # m_hWnd, HWND_TOPMOST, ..., SWP_NOSIZE | SWP_NOMOVE
                _user32.SetWindowPos(m_hWnd, -1, 0, 0, 0, 0, 0x0001 | 0x0002)
```

larsoner:
Marking for merge when green, thanks in advance @tomstoll !

PR #451, opened 2024-06-21
MAINT: Make sure CIs are still green

larsoner:
Let's see if things are still green here :crossed_fingers:


larsoner:
Okay there are some tricky bugs here on macos-14+ (arm64). I'm going to disable that for now and try adding it later probably after some local testing.

larsoner:
@nordme finally managed to get CIs green here. I think you were doing some `expyfun` testing. In any case, I'd suggest updating to latest `main` locally and hopefully things work better!

rkmaddox:
Thanks for all the maintenance @larsoner!

PR #491, opened 2025-03-07
Overdue commit for dual-monitor display fix

megadmin:
Note from SMB: These changes have been sitting on Newnin (newest stim computer in the I-LABS MEG lab) for some time. The line changes in __exp_controller_ were necessary for some dual-monitor and projector displays .  I think _sync_test_dualmon.py_ was adapted from the existing _sync_test.py_ at the time to test lag times with the new setup.

Per the requested merge: We only have the one computer with which to test the display functionality, but I can attest that the `screen.x` | `screen.y trick` was necessary and doesn't seem to conflict with other uses of ExpyFun.


larsoner:
Thanks @megadmin !

megadmin:
And thank you @larsoner.  In retrospect I should have pulled in the newer changes on _main_ first, before making the pull request, so thanks for taking care of that part.  I did need to `pip install` the h5io package, but after that the merged update works just fine on Newnin (at least for the current paradigms).

PR #522, opened 2025-09-16
compatibility with Pyglet 2+

drammock:
our pin to Pyglet < 1.6 comes with a pin (on conda-forge) to Python 3.5 or lower. It seems to be possible to ignore that by installing pyglet from PyPI, but it would be better if things just worked with newer Pyglet. This PR is hopefully pretty close already (almost all tests pass), but there's a pesky segfault when running `test_visuals.py` that I haven't cracked yet (partial traceback):

```
Current thread 0x0000758b0eeb8740 (most recent call first):
  File "/opt/mambaforge/envs/expyfun/lib/python3.13/site-packages/pyglet/graphics/vertexdomain.py", line 755 in draw_subset
  File "/opt/mambaforge/envs/expyfun/lib/python3.13/site-packages/pyglet/graphics/vertexdomain.py", line 136 in draw
  File "/opt/mambaforge/envs/expyfun/lib/python3.13/site-packages/pyglet/sprite.py", line 848 in draw
  File "/opt/expyfun/expyfun/visual/_visual.py", line 1128 in draw
```

- [x] update CIs to use newer pyglet
- [x] fix segfault
- [x] get rid of `_new_pyglet` and require Pyglet 1.4+
- [x] get rid of checks for OpenGL (assume it's there if testing is being done)
- [x] restore Pyglet 1.4 compat to allow for easier transition
- [x] fix TODO commented-out code bits
- [x] get CIs happy
- [x] fix primitive drawing on macOS
- [x] verify that CircleCI outputs match
- [x] check timing on experimental computers with `pyglet < 2` and `pyglet > 2`

Closes #440

### DIFF
--- a/expyfun/_tdt_controller.py
+++ b/expyfun/_tdt_controller.py
@@ -170,10 +170,8 @@ class TDTController(Keyboard):  # lgtm [py/missing-call-to-init]
                 self.rpcox = connect_rpcox(
                     name=use_model, interface=self.interface, device_id=1, address=None
                 )
-            except Exception as exp:
-                raise OSError(
-                    f"Could not connect to {self._model}, is it turned on?"
-                )
+            except Exception:
+                raise OSError(f"Could not connect to {self._model}, is it turned on?")
         else:
             msg = (
                 "TDT is in dummy mode. No sound or triggers will "


larsoner:
https://github.com/LABSN/expyfun/pull/440 is probably needed. Old vs new OpenGL shader and drawing methods will be involved I think. But maybe not!

larsoner:
... In other words, most of our native shape drawing stuff might need to be rewritten. Maybe we could get away with a dependency on something else like vispy for this nowadays.

larsoner:
Okay locally with Pyglet 2.1.1 I can at least `python examples/basic_experiment.py` and some stuff plays and displays. I'll update the TODO list above

larsoner:
Ugh... having spent some hours on this, there are two tough problems:

1. macOS-arm64 isn't working. I think the OpenGL context isn't being created properly before we try to draw to it. I've tried forcing `visible=True` in tests, which didn't seem to help. Using `full_screen=True` I get a little bit farther but hit a different error. I am not sure how we can get around this. It seems to have to do with differences in event processing. A nuclear option would be to start using PySide6 for window and event management rather than Pyglet. For now... maybe we have to disable macOS testing :scream: 

2. Advanced text segfaults when I put `draw=True` back in. I think this is a Pyglet bug. I need to whittle it down to one I think.

drammock:
this is great progress. I'll see what I can do to help next week (kinda swamped today & tomorrow)

larsoner:
@drammock I think I'll make a separate issue to fix drawing on macOS. I think it'll take a bit of investigating. In the meantime I think we need to check that stimulus timing is still good somehow, but I think the MEG center folks will be busy. So not sure what to do there

drammock:
> In the meantime I think we need to check that stimulus timing is still good somehow

is that something that @rkmaddox could help out with?  We could also maybe ask @mwinn83, he's been using Expyfun in his lab recently.

expyfun/_utils.py
```diff
@@ -432,45 +421,17 @@ def verbose_dec(function, *args, **kwargs):
...
-        else:
-            if raise_error:
-                print("Found AVbinSource for old Pyglet 2")
-    if raise_error and not good:
-        raise RuntimeError(
-            "Video support not enabled, got exception(s):\n"
-            "\n***********************\n".join(exceptions)
-        )
+        if raise_error:
+            print("Found FFmpegSource for new Pyglet")
```
drammock, line 434:
```suggestion
            print("Found FFmpegSource for Pyglet")
```

expyfun/visual/_visual.py
```diff
@@ -1201,8 +1237,7 @@ class Video:
     Notes
     -----
     This is a somewhat pared-down implementation of video playback. Looping is
-    not available, and the audio stream from the video file is only available when
-    using the sound card as audio controller and ``"pyglet"`` as the sound card backend.
+    not available, and the audio stream from the video file is discarded.
```
drammock, line 1246: this looks like a messed-up rebase?

expyfun/visual/_visual.py
```diff
@@ -1213,32 +1248,32 @@ def __init__(
...
+        except Exception as exc:
+            warnings.warn(
+                "FFmpeg decoder could not be instantiated, decoding "
+                f"performance could be compromised:\n{exc}"
+            )
         self._source = load(file_name, decoder=decoder)
         self._player = Player()
         with warnings.catch_warnings(record=True):  # deprecated eos_action
             self._player.queue(self._source)
+        self._player._audio_player = None
```
drammock, line 1282: ditto (rebase snafu)

drammock:
> Should be good to go other than needing a sync test, ideally on Pyglet < 2 and Pyglet >= 2.

amazing!  early next week I'll try to work up some simple instructions / scripts to facilitate setting up that testing, to make things as easy as possible for our testers.

larsoner:
Okay, verified on Windows 10 with picoscope + parallel port triggering + TDT RM1 + photodiode with:
```
SCREEN_NUM=0 AUDIO_CONTROLLER=tdt TDT_MODEL=RM1 TRIGGER_CONTROLLER=parallel TRIGGER_ADDRESS=0xDCF8 python -u examples/sync/sync_test.py
```
I get pretty reliable ~32ms delay, which seems to be the best I can get. Can't seem to get it down to 16ms 🤷 . But performance seems to be the same on this PR with Pyglet 2.1 as it is on `main` with Pyglet 1.5.

I'll see if I can also replicate with parallel+video on Linux just for fun, but I won't bother with the audio output there. In the meantime @drammock I think we can merge, agreed?

larsoner:
Okay the timing was bad on Linux on my machine for both old pyglet+main and new pyglet+this PR. Behaviors were very similar, including erratic 16 or 32ms delays, sometimes the white wouldn't even flash, etc. So at least things aren't worse here on Linux.

Also I forgot to post the Picoscope image for Windows above, TDT in blue with photodiode in red, triggered to the parallel port high:
20250922_latency

drammock:
locally (Xubuntu) with rtmixer I get approx 50/50 split between 16-17ms IFI and 32-34ms IFI. I didn't notice any missed flashes.  If I request pyglet backend I get a larger proportion of 32-34ms IFIs, and several missed flashes.  Which I agree, is not great but not worse than it was with Pyglet <1.6. So yeah, I guess let's merge, since worst case we always have the `download_version` mechanism to pin experiments.

larsoner:
Just to close the loop on this... @NeuroLaunch tested this code at ILABS yesterday and things appear reliable timing-wise! :tada:

PR #525, opened 2025-09-17
MAINT: Fix Windows

larsoner:


larsoner:
Also now requires 3.10 and tests on macos-latest instead of deprecated macos-13. Hopefully that comes back green :crossed_fingers:

larsoner:
... okay macos-13 to macos-latest will need to wait for Pyglet to be newer

drammock:
Locally I was seeing considerable flakiness too...  `test_play_sound` was passing when I only ran that test with `-k`, but failing when I ran the full test suite.  `test_ec` would also sometimes fail, sometimes pass, sometimes segfault

larsoner:
I also reverted the 3.10 stuff. That's a pain without newer Pyglet. So this just hacks around / `xfail`s the Windows experiment_controller test for now

drammock (approved):

PR #526, opened 2025-09-24
MAINT: Speed up flip

larsoner:
I realized that in #522 I made it create a new `Rectangle` object every time the window is flipped. There is no reason to do this, and it will cause the visual shaders to be recompiled over and over (unless the drivers do some smart caching, which they might). So this now creates and draws it first in the window setup function, then reuses it every flip.


larsoner:
No I don't see any differences

larsoner:
By that I mean no I didn't test too much (like not with a scope)... on macOS and Linux behavior seemed the same in terms of inter-flip intervals etc

PR #529, opened 2025-09-26
MAINT: Remove version downloading utils

larsoner:


larsoner:
FYI @drammock I manually locally checked out our ancient (2014!) `2.0.0` tag, made some tiny tweaks to make it so I could `python setup.py sdist bidst_wheel` it, and uploaded it to PyPI. That allowed me to reserve the name and set up trusted publishing, which this PR also implements with `release.yml`

larsoner:
I'll self-merge so that #494 can proceed without `setup.py` errors, but @drammock feel free to review when you get a chance later on and I'm happy to open a follow-up

